PR #60, opened 2019-08-28
!): Fix for ubuntu 18.04 link error.

tbfly:
Fix for Linux ubuntu 18.04 desktop bazel run example link error.


tbfly:
> Thanks for the PR.
> 
> If you append `--define 'libunwind=true'` to the bazel build command, it should solve the link error as well. We don't do this by default because the define flag is not needed if we start from a fresh ubuntu 18.04 system. We will document this issue in the coming release.

thx

PR #138, opened 2019-09-28
log the errors always

jiqimaogou:
> constexpr int kMaxNumAccumulatedErrors = 1000;

only the errors count is large than 1000, the errors can be printed.

I think error log is very important to developers. It should be printed when it happened.


jiqimaogou:
@googlebot I signed it!

hadon:
Thanks jiqimaogou.  I believe you can see these errors immediately using the command line option either "--v=2" or "--vmodule=calculator_graph=2".  Let us know if these options are not sufficient for your use case.

PR #172, opened 2019-10-14
some documentation typos

fenollp:
Sorry for the noise! I'll understand if you don't merge this, was just a it annoying.


chuoling:
Thanks for reporting this, and sorry for the really late response.

The discrepancies in code comments (i.e., options v.s. node_options, and mediapipe.XOptions v.s. type.googleapis.com/mediapipe.XOptions) come from the different syntax we have in google-internal code base v.s. external OSS code. While for open sourcing we are able to auto-convert the syntax in the graph representation (i..e, pbtxt) files, automating such conversion in code comments can be trickier and we currently don't have plans to address them.

Issue #88, opened 2019-09-06
How to build mediapipe as standalone aar library ?


aminm2a:
What is the steps to do to build mediapipe as a standalone aar library to be integrated in a project ?

ZhangWen0301:
same problem

lqz:
same problem

mgyong:
We will be looking into this

mgyong:
@aminm2a @lqz @ZhangWen0301 
Pls note in [MediaPipe v0.6.3](https://github.com/google/mediapipe/releases/tag/v0.6.3) we released support for [building AAR](https://github.com/google/mediapipe/blob/master/mediapipe/docs/android_archive_library.md)

billhung:
The instruction said to "put the following code into mediapipe/examples/android/src/java/com/google/mediapipe/apps/aar_example/BUILD"

But I looked in v0.6.3 linked above, and there's no aar_example in the said folder.
I don't know how to proceed, can someone comment on what I am missing?

lqz:
> @aminm2a @lqz @ZhangWen0301
> Pls note in [MediaPipe v0.6.3](https://github.com/google/mediapipe/releases/tag/v0.6.3) we released support for [building AAR](https://github.com/google/mediapipe/blob/master/mediapipe/docs/android_archive_library.md)

Can there be more help documents, such as how to use the interface？

gordinmitya:
```bazel
load(“//mediapipe/java/com/google/mediapipe:mediapipe_aar.bzl”, “mediapipe_aar”)
mediapipe_aar(
name = “mp_object_detection_aar”,
calculators = [“//mediapipe/graphs/object_detection:mobile_calculators”],
)
```

https://medium.com/@doquocbao_91868/build-object-detection-mediapipe-aar-and-integrate-in-smart-mirror-running-on-careos-c074db846313

PR #181, opened 2019-10-16
Made AndroidStudio Project for HandTrackingGpu

alizahidraja:


jiuqiant:
Thanks for the pull request.

We probably can't accept this PR for two reasons:
- It's not a good solution to copy the framework's java and proto source code to an app directory. If we accept such solution, the java and proto source code will be duplicated in every single app directory.
- To fix https://github.com/google/mediapipe/issues/88, MediaPipe team is going to release a Bazel rule to generate an AAR file that can be imported into Android Studio. With the AAR, you can use Mediapipe in your own Android project.

Just for preview, we will provide a mediapipe_aar() Bazel rule in the next release. The users need to define a mediapipe_aar() rule in a BUILD file within the MediaPipe repo. For example,
putting the following code into mediapipe/examples/android/aar_demo/BUILD.
```
mediapipe_aar(
    name = "my_aar",
    calculators = ["//mediapipe/graphs/face_detection:mobile_calculators"],
)
```

Then, run the following Bazel command to generate the AAR.

```
$ bazel build -c opt --fat_apk_cpu=arm64-v8a,armeabi-v7a mediapipe/examples/android/aar_demo:my_aar
```

You will get

```
Target //mediapipe/examples/android/src/java/com/google/mediapipe/apps/aar_demo:my_aar up-to-date:
  bazel-genfiles/mediapipe/examples/android/src/java/com/google/mediapipe/apps/aar_demo/my_aar.aar
```
Finally, import this AAR file into Android Studio.

We believe this is a better solution to integrate MediaPipe into Android Studio.

alizahidraja:
I understand
But I guess you can merge the java and protoc files as not everyone have easy access to those if they want a android studio project. Or we can have a common directory for all the common java and protoc files and the app build gradles too along with each project 
This will help the user and save a lot of their time too

PR #183, opened 2019-10-18
add RectsPulsarCalculator and NormalizedRectsPulsarCalculator

fenollp:
This calculator turns a `vector<T,N>` packet into `N` packets of `T`.

Tests are coming.

### DIFF
--- a/mediapipe/calculators/core/pulsar_calculator.h
+++ b/mediapipe/calculators/core/pulsar_calculator.h
@@ -1,0 +1,58 @@
+// Copyright 2019 The MediaPipe Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


mediapipe/calculators/core/pulsar_calculator.h
```diff
@@ -0,0 +1,58 @@
...
+    cc->SetOffset(TimestampDiff(0));
+    return ::mediapipe::OkStatus();
+  }
+
+  ::mediapipe::Status Process(CalculatorContext* cc) final {
+    const auto& items = cc->Inputs().Index(0).Get<std::vector<T>>();
+    auto ts = std::max(cc->InputTimestamp(), ts_);
+    for (const auto& item : items) {
+      auto just_one = absl::make_unique<T>(item);
+      cc->Outputs().Index(0).Add(just_one.release(), ts++);
```
fenollp: > ts++

This turns an input packet at timestamp `t` of `N` things into `N` packets each at timestamp `t`, `t+1`, ..., `t+N` thus potentially messing up downstream calculations (output packets won't be matched to the intended timestamps).
This calculator was intended as a work around the fact that the current `ImageCroppingCalculator` doesn't take vector packets. Any reason why?

Do you guys have the intention of adding some `MapCalculator` that would apply a subgraph onto a packet of vector?

Do you have any advice on how I could generate multiple packets at the same timestamp?
Thanks @mgyong

fenollp:
Actually, today's `SplitVectorCalculator` is enough for my needs.

I am still curious about how you think a `MapCalculator` could be done (applying some calculator to `N` streams of inputs, `N` being known only at runtime. With a side-packet concurrency granularity parameter). 
Can one add input/output streams and calculators instances to an already `Initialized` graph?

PR #229, opened 2019-11-08
enable NNAPI in CPU path

freedomtan:
With NNAPI delegate added, Android devices with floating point
NNAPI device(s) could run facedectioncpu and objectdectioncpu
accelerated. On devices without accelerators, tflite interpreter
will simply fall back to CPU.


freedomtan:
Similar code is in recent update. Slose this. https://github.com/google/mediapipe/blob/master/mediapipe/calculators/tflite/tflite_inference_calculator.cc#L523-L530

PR #319, opened 2019-12-13
Fix typo in iOS Setup documentation

lukereichold:


chuoling:
Thank you for reporting this, and sorry for the really late response. Verified that the typo is no longer in the latest documentation.

PR #412, opened 2020-01-21
Update install.md

razak-io:
"linux_ffmpeg" needs to be changed as well as "linux_opencv" for OpenCV 4


jiuqiant:

PR #432, opened 2020-02-04
Correct link to source pbtxt file

jeroenboeye:
The url pointed to the object detection instead of tracking version


jeroenboeye:
@googlebot I signed it!

chuoling:
Thanks for reporting this. The link is now correct in the revamped doc in https://github.com/google/mediapipe/blob/master/docs/solutions/box_tracking.md.

PR #553, opened 2020-03-25
Update install.md

yingshaoxo:
Add a JDK installation command for the running of an android app.


yingshaoxo:
@googlebot I signed it!

PR #633, opened 2020-04-21
Fix bad link and update graph text in hello_world_android.md.

H1Gdev:


chuoling:
Thanks for reporting this. Merged into our internal code base, and will be pushed to GitHub in the next release.

PR #678, opened 2020-05-03
documentation fixes, perform all examples builds/runs in Dockerfile

allenday:
the build is slow as I introduce a dependency on downloading BigBuckBunny - perhaps better to add a small video file to the repo that has audio (object tracking video already present does not, and this causes failures).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:latest
+FROM ubuntu:18.04
 
 MAINTAINER <mediapipe@google.com>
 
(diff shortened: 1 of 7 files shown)


tslater:
I get an error trying to build your dockerfile:
```Step 18/40 : RUN bazel build -c opt     --define MEDIAPIPE_DISABLE_GPU=1     --define no_aws_support=true     --linkopt=-s     mediapipe/examples/desktop/object_detection:object_detection_tensorflow
 ---> Running in 165462de4be8
Extracting Bazel installation...
ERROR: The 'build' command is only supported from within a workspace (below a directory having a WORKSPACE file).
See documentation at https://docs.bazel.build/versions/master/build-ref.html#workspace
The command '/bin/sh -c bazel build -c opt     --define MEDIAPIPE_DISABLE_GPU=1     --define no_aws_support=true     --linkopt=-s     mediapipe/examples/desktop/object_detection:object_detection_tensorflow' returned a non-zero code: 2```

jiuqiant:
Allen, most of the changes of this PR were implemented in https://github.com/google/mediapipe/commit/b133b0f2004c831c2be004291d3fcf7a378a48f4.
Building and running all examples might be too heavy for some people who only need a clean environment. We can put them as comments. What do you think?

allenday:
Yes, good idea. Shall I revise the PR here or we do it internally?

PR #705, opened 2020-05-07
unusual gap

keshavgbpecdelhi:
Two tab =》 One tab
Or it may be like python 2 or python 3


chuoling:
Thanks for reporting this. Merged into our internal code base, and will be pushed to GitHub in the next release.

PR #757, opened 2020-06-02
In normal case channel order is NHWC

brucechou1983:
The original setting will result in error when input height != width


mcclanahoochie:
Hi,
Thanks for your contribution.
This change is not strictly necessary, as the ordering of options defined in .proto files (and their [field ID numbers](https://developers.google.com/protocol-buffers/docs/overview#simple)) doesn't impact how they are used or specified in a MediaPipe graph (.pbtxt).
Re-assigning the field ID numbers here will not change the ordering, and it may actually break existing graphs. 
Please see [here](https://developers.google.com/protocol-buffers/docs/overview#simple) for more info about how .proto files are defined (and the reason for unique ID numbers).
If you find there is something not working with your segmentation mode (when width!=height), and swapping these values in your graph works, then there may be an issue elsewhere in the tensors to segmentation calculator, but it would not be fixed here.
Thanks,

PR #790, opened 2020-06-11
Fix typo in comment.

H1Gdev:
There are many `iff` in comments.
https://github.com/google/mediapipe/search?q=iff

If it is a comment rule, just close it.


chuoling:
Thanks for reporting this. `iff` is actually used in those comments as an abbreviation for `if and only if`.

PR #895, opened 2020-07-09
Support for OpenCV on Ubuntu 20.04 in opencv_linux.BUILD

fenollp:
OpenCV paths changed a little on latest LTS release.
This change makes builds pass on a fresh install of Ubuntu 20.04.


chuoling:
Thanks for reporting this. The change has been merged into our internal code base upstream, and will be pushed to GitHub in the next release.

PR #901, opened 2020-07-12
Fix spacing for README.md under android examples.

ucalyptus:
Just a small fix.
cc: @jiuqiant @chuoling


chuoling:

PR #909, opened 2020-07-15
Fix sample code in doc.

H1Gdev:


chuoling:

PR #938, opened 2020-07-24
Add option for cuda optflow.

cheoljunlee:
Current version of mediasequence does not support gpu for optical flow.
So, I added option for that.


cheoljunlee:
@googlebot I signed it!

cheoljunlee:
@googlebot I fixed it.

Issue #1033, opened 2020-08-22
iris_depth_from_image_desktop example hangs


shortcipher3:
I am using the provided Dockerfile and add the following line to build the iris depth example.

```
bazel build -c opt --define MEDIAPIPE_DISABLE_GPU=1 mediapipe/examples/desktop/iris_tracking:iris_depth_from_image_desktop
```

It built fine.  then from the docker image I ran:
```
GLOG_logtostderr=1 bazel-bin/mediapipe/examples/desktop/iris_tracking/iris_depth_from_image_desktop --input_image_path=<input image path> --output_image_path=<output image path>
```
Replacing the image paths with my desired images.

I get some output then it hangs:
```
I20200822 21:16:33.688568    53 iris_depth_from_image_desktop.cc:130] Get calculator graph config contents: # MediaPipe graph that performs iris distance computation on desktop with
# TensorFlow Lite on CPU.
# Used in the example in
# mediapipie/examples/desktop/iris_tracking:iris_depth_from_image_desktop.
...
# Renders annotations and overlays them on top of the input images.
node {
  calculator: "IrisAndDepthRendererCpu"
  input_stream: "IMAGE:input_image"
  input_stream: "FACE_LANDMARKS:face_landmarks"
  input_stream: "EYE_LANDMARKS_LEFT:left_eye_contour_landmarks"
  input_stream: "EYE_LANDMARKS_RIGHT:right_eye_contour_landmarks"
  input_stream: "IRIS_LANDMARKS_LEFT:left_iris_landmarks"
  input_stream: "IRIS_LANDMARKS_RIGHT:right_iris_landmarks"
  input_stream: "NORM_RECT:face_rect"
  input_stream: "LEFT_EYE_RECT:left_eye_rect_from_landmarks"
  input_stream: "RIGHT_EYE_RECT:right_eye_rect_from_landmarks"
  input_stream: "DETECTIONS:face_detections"
  input_side_packet: "IMAGE_FILE_PROPERTIES:image_file_properties"
  output_stream: "IMAGE:output_image"
  output_stream: "LEFT_IRIS_DEPTH_MM:left_iris_depth_mm"
  output_stream: "RIGHT_IRIS_DEPTH_MM:right_iris_depth_mm"
}
I20200822 21:16:33.690336    53 iris_depth_from_image_desktop.cc:136] Initialize the calculator graph.
I20200822 21:16:33.700639    53 iris_depth_from_image_desktop.cc:59] Load the image.
[...]

PR #1098, opened 2020-09-09
Bug iris no face detected hangs

shortcipher3:
I opened an issue: https://github.com/google/mediapipe/issues/1033 for a bug I found where the iris depth detection can hang on an individual image.  One case this happens on is when no faces are detected.  My proposed fix is to throw an error when an option is provided that requires the detections vector for non-max suppression not be empty.


shortcipher3:
@googlebot I fixed it

shortcipher3:
@googlebot I fixed it.

PR #1102, opened 2020-09-10
Bug iris no face detected hangs

shortcipher3:
I opened an issue: #1033 for a bug I found where the iris depth detection can hang on an individual image. One case this happens is when no faces are detected. My proposed fix is to throw an error when an option is provided that requires the detections vector for non-max suppression not be empty.


DandyLorenz:
> 我打开了一个问题：[＃1033](https://github.com/google/mediapipe/issues/1033)，是一个错误，我发现虹膜深度检测可以挂在单个图像上。发生这种情况的一种情况是没有检测到脸部。我建议的解决方法是，在提供要求非最大抑制的检测向量不为空的选项时引发错误。

PR #1143, opened 2020-09-25
Fix FilledRoundedRect render in annotation_render.cc

aaaron7:
cv::line wont accept thickness less than 0 but DrawFilledRoundedRectangle always pass -1 for thickness.


aaaron7:
@googlebot I signed it!

PR #1161, opened 2020-09-29
[fix] Just a typo (resulTing)

roj4s:
Signed-off-by: Luis Miguel Rojas Aguilera <rojas@icomp.ufam.edu.br>


roj4s:
@googlebot I signed it!

PR #1171, opened 2020-10-01
:pencil: add how to python in face_mesh model

AlissonSteffens:
So, I added this, to people who want to use face_mesh model in python. I don't know why it wasn't here, maybe you don't want this kind of tutorial.
BTW, If you think it is good to have, I just added. a Python code to run 'face_landmark.tflite' in Python.


AlissonSteffens:
@googlebot I signed it!

jiuqiant:
Thanks for reporting this. We have MediaPipe FaceMesh Python solution now. Please take a look. Thanks.

PR #1254, opened 2020-11-05
fix DetectionUniqueIdCalculator consume prob

mapleyuan:
The subgraph 'ObjectTrackingSubgraphGpu' is useful in many cases. In some cases we need to connect the input stream 'new_detections' for other calculators, but the ‘DetectionUniqueIdCalculator’ can't accept it. I think use 'ConsumeOrCopy' will be better.

image


mapleyuan:
@googlebot I signed it!

mapleyuan:
@googlebot I fixed it.

PR #1256, opened 2020-11-05
Make object tracking works on iOS

aaaron7:
The problem is MPPGraph.mm uses frame sequential number as mediapipe timestamp which makes time sensitive logic failed in lots of calculators, such as PacketResamplerCalculator, BoxTrackerCalculator. 

The object tracking graph works perfectly on iOS after correcting timestamp to current microseconds.

### DIFF
--- a/mediapipe/examples/ios/common/CommonViewController.h
+++ b/mediapipe/examples/ios/common/CommonViewController.h
@@ -18,6 +18,7 @@
 #import "mediapipe/objc/MPPGraph.h"
 #import "mediapipe/objc/MPPLayerRenderer.h"
 #import "mediapipe/objc/MPPPlayerInputSource.h"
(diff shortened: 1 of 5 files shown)


camillol:
Thanks for the bug report! I think you've identified the right issue. To fix it, I would suggest changing `mediapipe/examples/ios/common/CommonViewController.mm` to call sendPixelBuffer:intoStream:packetType:timestamp:, passing the timestamp it received from the video source. That would also align with the behavior of the corresponding code on Android. @eknight7 could you take a look?

aaaron7:
@camillol Good suggestion! I've updated the code by using MPPTimestampConverter directly convert video CMTime to Timestamp.

chuoling:
The change suggested above by @camillol has been incorporated into our upstream source, and will be pushed in the next release.

PR #1261, opened 2020-11-06
Update pose docs

Pangoraw:
We need to save the `annotated_image` instead of the `image` to see the drawings in the saved file.


DandyLorenz:

PR #1289, opened 2020-11-12
Implement create Vectors method in PacketCreator class.

H1Gdev:
- createInt16Vector
- createInt32Vector
- createInt64Vector
- createFloat64Vector


H1Gdev:
`Array`s also needs to send size, so I want you to implement `Vector`s.

PR #1395, opened 2020-12-15
model

MaRauder111:
hand landmark model and palm detection model is not properly link.


DandyLorenz:
> hand landmark model and palm detection model is not properly link.

PR #1413, opened 2020-12-20
fix typo (with -> width)

homuler:
Fix typo (with -> width) in error messages.


homuler:
Closes because it's fixed at v0.8.3.1.

PR #1542, opened 2021-01-24
make box_tracker.cc more portable

homuler:
We cannot compile `box_tracker.cc` on Windows because `usleep` is not defined.
I updated so that `absl::SleepFor` is called instead.


homuler:
Closes because it's fixed by v0.8.3.

Issue #1575, opened 2021-02-02
Draw landmarks for upper-body pose tries to access not-existing landmarks.


cansik:
The current sample code in python is handing over all of the pose connections, regardless if in the upper-body mode or not:

```python
# at the beginning
pose = mp_pose.Pose(
    upper_body_only=True,
    min_detection_confidence=0.5, 
    min_tracking_confidence=0.5)

# in draw loop
mp_drawing.draw_landmarks(image, results.pose_landmarks, mp_pose.POSE_CONNECTIONS)
```

This results in a Landmark index out of range error:

```
ValueError: Landmark index is out of range. Invalid connection from landmark #30 to landmark #32.
```

What we would need to have is either a more intelligent `draw_landmark` method which does not connect landmarks that do not exist, or a second frozen connection list for the upper body pose. I have already created one:

```python
UPPER_POSE_CONNECTIONS = frozenset([
    (PoseLandmark.NOSE, PoseLandmark.RIGHT_EYE_INNER),
    (PoseLandmark.RIGHT_EYE_INNER, PoseLandmark.RIGHT_EYE),
    (PoseLandmark.RIGHT_EYE, PoseLandmark.RIGHT_EYE_OUTER),
    (PoseLandmark.RIGHT_EYE_OUTER, PoseLandmark.RIGHT_EAR),
    (PoseLandmark.NOSE, PoseLandmark.LEFT_EYE_INNER),
    (PoseLandmark.LEFT_EYE_INNER, PoseLandmark.LEFT_EYE),
    (PoseLandmark.LEFT_EYE, PoseLandmark.LEFT_EYE_OUTER),
    (PoseLandmark.LEFT_EYE_OUTER, PoseLandmark.LEFT_EAR),
    (PoseLandmark.MOUTH_RIGHT, PoseLandmark.MOUTH_LEFT),
    (PoseLandmark.RIGHT_SHOULDER, PoseLandmark.LEFT_SHOULDER),
    (PoseLandmark.RIGHT_SHOULDER, PoseLandmark.RIGHT_ELBOW),
    (PoseLandmark.RIGHT_ELBOW, PoseLandmark.RIGHT_WRIST),
    (PoseLandmark.RIGHT_WRIST, PoseLandmark.RIGHT_PINKY),
    (PoseLandmark.RIGHT_WRIST, PoseLandmark.RIGHT_INDEX),
    (PoseLandmark.RIGHT_WRIST, PoseLandmark.RIGHT_THUMB),
[...]

PR #1576, opened 2021-02-02
Added missing upper-pose-connection set

cansik:
As already described in #1575 , the upper-pose connections are missing and using the pose model with upper-body only results in an out of range error.

This PR adds a new connection set `UPPER_POSE_CONNECTIONS` only containing the upper pose connections. In my opinion it makes more sense to add such a connection set, instead of changing the `draw_landmarks()` method (which already check if landmarks are available).

Would be great if you merge this PR, that you also make a hint on the [manual page](https://google.github.io/mediapipe/solutions/pose.html) of the pose estimation.

### DIFF
--- a/mediapipe/python/solutions/pose.py
+++ b/mediapipe/python/solutions/pose.py
@@ -41,7 +41,7 @@ from mediapipe.python.solution_base import SolutionBase
 
 
 class PoseLandmark(enum.IntEnum):
-  """The 25 (upper-body) pose landmarks."""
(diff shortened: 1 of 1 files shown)


cansik:
@googlebot I signed it!

Preetham-Yeddula:
How to pass UPPER_POSE_CONNECTIONS set to draw_landmarks()?
Thanks in advance.

jiuqiant:
Thanks for reporting this. We added mp_pose.UPPER_BODY_POSE_CONNECTIONS for upper body use case.

PR #1652, opened 2021-02-23
Bump protobuf to 3.15.1 to work around Bazel 4 incompatibility

johnboiles:
I ran into [this issue](https://github.com/bazelbuild/bazel/issues/12887) with building MediaPipe using Bazel 4.0.0. Apparently you need to use at least Protobuf 3.13 with Bazel 4.0.0.

This PR bumps Protobuf to 3.15.1.

I'm new to MediaPipe and Bazel. I'm not 100% sure I generated `third_party/com_google_protobuf_fixes.diff` correctly but it seems to work for me.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -125,16 +125,16 @@ http_archive(
 # ...but the Java download is currently broken, so we use the "source" download.
 http_archive(
     name = "com_google_protobuf_javalite",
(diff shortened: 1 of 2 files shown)


johnboiles:
@googlebot I signed it!

johnboiles:
Looks like [rules_apple](https://github.com/bazelbuild/rules_apple/releases) is only tested with Bazel 3.7 so that might also block fully moving to Bazel 4 until that's solved.

mirh:
I think this can be closed.

PR #1676, opened 2021-02-28
build objectron_detection_3d:mobile_calculators for iOS

homuler:
This PR enables to build [object_detection_3d:mobile_calculators](https://github.com/google/mediapipe/blob/master/mediapipe/graphs/object_detection_3d/BUILD#L29) for iOS.

I referred to the following code.
https://github.com/google/mediapipe/blob/350fbb2100ad531bc110b93aaea23d96af5a5064/mediapipe/calculators/core/BUILD#L240-L248

I think other calculators that depends on `tensor_util` (e.g. `tflite_tensors_to_objects_calculator`) would also require the same options, but I leave them alone for now because I could not find where they are used.

### DIFF
--- a/mediapipe/modules/objectron/calculators/BUILD
+++ b/mediapipe/modules/objectron/calculators/BUILD
@@ -181,6 +181,15 @@ cc_library(
     hdrs = [
         "tensor_util.h",
     ],
+    copts = select({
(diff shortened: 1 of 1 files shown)


ahmadyan:
Objectron required Android for rendering in the later stage in the pipeline, at the moment apple targets are not supported.

PR #1716, opened 2021-03-07
fix: build inference_calculator_gl for android

homuler:
`inference_calculator_gl` includes header files under `mediapipe/util/android/file/base`, but those files are not declared in `deps`.
https://github.com/google/mediapipe/blob/350fbb2100ad531bc110b93aaea23d96af5a5064/mediapipe/calculators/tensor/inference_calculator_gl.cc#L32-L36

This PR adds those headers to `deps` explicitly.


homuler:
Fixed by https://github.com/google/mediapipe/commit/63e679d99ca45b30514a9d84c9351a2d77bb9ba0#diff-3f12815451147226887c6cb787adfd7fd289e5e04162147e9127fd02b8879486

PR #1759, opened 2021-03-17
Fixed inconsistent type in contract and open of input side packets

libraua:
Inside ThresholdingCalculator::GetContract we are saying that it will be double but in actual open we expecting float. This fix just fixing this inconsistency.


libraua:
@googlebot I signed it!

jiuqiant:
Fixed in https://github.com/google/mediapipe/commit/a9b643e0f5978425d948050a2228398929733d05.

Issue #1733, opened 2021-03-11
Mediapipe build error


studido:
Hey there,
I'm attempting to build mediapipe on a jetson nano and am running into the following build error. It concerns the image_to_tensor_converter_opencv.cc file:

(line) return tensor;
could not convert 'tensor' from 'mediapipe::Tensor' to 'absl::lts_2020_09_23::StatusOrmediapipe::Tensor'

I tried adding a direct conversion in the form of:
return absl::lts_2020_09_23::StatusOrmediapipe::Tensor(std::move(tensor));
but I get a different error saying absl::lts_2020_09_23::StatusOrmediapipe is not declared.

Any help of ideas of what could be causing this would be greatly appreciated!

sgowroji:
@studido Could you please share  complete details of the above issue , like what solutions and platform are you aiming for. And also share the detail build error.

milad-4274:
I faced similar errorsince last update. all examples cpu on ubuntu.

studido:
@sgowroji 
Thanks for your reply.

I'm trying to build mediapipe in python for aarch64 (ubuntu 18.04) on an nvidia jetson nano. 

Currently have been trying to follow this guide:
https://github.com/jiuqiant/mediapipe_python_aarch64

Following those steps I removed the OpenCV dependencies then tried to build a wheel for python with:
python3 setup.py gen_protos && python3 setup.py bdist_wheel

It uses bazel to build a wheel out of the mediapipe repo but in the later stages of the build I get faced with the C++ compilation error mentioned in my post. 
I attached two screenshots of the build error, one with the original error and a second with the only solution I attempted to try. Instead of returning tensor which caused the conversion error I tried to return absl::lts_2020_09_23::StatusOrmediapipe::Tensor(std::move(tensor)) directly but that produced an is not declared error. I couldn't find anymore resources online about this error and am unsure how to proceed. 
Thank you for your help!

![159942583_1074407583038608_3706416316026448497_n](https://user-images.githubusercontent.com/49419297/111005118-0f6e5780-833f-11eb-9bcc-a4524be38eb5.png)
![159512937_315151003274733_1460369004166364305_n](https://user-images.githubusercontent.com/49419297/111005121-1006ee00-833f-11eb-833e-ea036911a253.png)

studido:
@milad-4274 Thanks, I will try it with an older release (0.8.0)

Emad-Eldeen:
[...]

PR #1867, opened 2021-04-10
Use gcc-8 and g++-8 in the Dockerfile

johnboiles:
This appears to be necessary since `0.8.3.1`.

See also https://github.com/google/mediapipe/issues/1733


jiuqiant:

johnboiles:
Thank you!

jiuqiant:
Fixed in https://github.com/google/mediapipe/commit/a9b643e0f5978425d948050a2228398929733d05.

PR #1969, opened 2021-05-02
Documentation Fix: Missing mp_holistic in PoseEstimation example

chr0nikler:
Seems like the var `mp_holistic` is referenced without being instantiated. Also looks like the preference is to create `mp_x` var rather than `mp.solutions.holistic` inline, so done at the top with the others.


chr0nikler:
@googlebot I signed it!

jiuqiant:

jiuqiant:
Fixed in https://github.com/google/mediapipe/commit/6abec128edd6d037e1a988605a59957c22f1e967.

PR #2026, opened 2021-05-13
Create stale.yml

rthadur:
As part of build automation , we are adding this file to improve nagging/stale issues which are marked stale after 7 days and closed after 14 days if there is no response from user.

cc @chanshah


jiuqiant:
Already added in a previous version.

PR #2027, opened 2021-05-13
Create bot_config.yml

rthadur:
As part of build automation , we are adding this file to auto assign issues to eng

cc @chanshah


jiuqiant:
Already added in a previous version.

PR #2039, opened 2021-05-16
Update pose.md

Abduttayyeb:
Corrected A Typo on line `190` , from **lanmark** to **landmark**


jiuqiant:
Thanks for bringing this issue to us. This is already fixed in the current version.

Issue #1889, opened 2021-04-14
AAR for MediaPipe Face Detection failed.


thunderbirdtr:
Hello, I was trying to build AAR file for learning/my project so I followed this link and it failed and has error 

https://google.github.io/mediapipe/getting_started/android_archive_library.html

I checked other issues saw protobuf patches but already in there (0.8.3-2) then I looked for that other issues but didn't see something similar.

`bazelisk build -c opt --host_crosstool_top=@bazel_tools//tools/cpp:toolchain --fat_apk_cpu=arm64-v8a,armeabi-v7a --strip=ALWAYS mediapipe/examples/android/src/java/com/google/mediapipe/apps/facedetectioncpu_aar:mp_face_detectioncpu_aar
`
```

DEBUG: Rule 'rules_foreign_cc' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "c2cdcf55ffaf49366725639e45dedd449b8c3fe22b54e31625eb80ce3a240f1e"
DEBUG: Repository rules_foreign_cc instantiated at:
  /home/thunderbirdtr/Projects/MpProjects/mediapipe-0.8.3.2/WORKSPACE:42:13: in <toplevel>
Repository rule http_archive defined at:
  /home/thunderbirdtr/.cache/bazel/_bazel_thunderbirdtr/47aecfbb39c3d4019351abcfc511078c/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
DEBUG: /home/thunderbirdtr/.cache/bazel/_bazel_thunderbirdtr/47aecfbb39c3d4019351abcfc511078c/external/org_tensorflow/third_party/repo.bzl:108:14: 
Warning: skipping import of repository 'com_google_absl' because it already exists.
DEBUG: /home/thunderbirdtr/.cache/bazel/_bazel_thunderbirdtr/47aecfbb39c3d4019351abcfc511078c/external/org_tensorflow/third_party/repo.bzl:108:14: 
Warning: skipping import of repository 'com_google_protobuf' because it already exists.
DEBUG: /home/thunderbirdtr/.cache/bazel/_bazel_thunderbirdtr/47aecfbb39c3d4019351abcfc511078c/external/org_tensorflow/third_party/repo.bzl:108:14: 
Warning: skipping import of repository 'com_google_googletest' because it already exists.
[...]

PR #2045, opened 2021-05-18
Updating ABSL cpp library into new LTS version

thunderbirdtr:
This change makes absl cpp library compatible with gcc 11 [1] in bazel compile.

Related with : https://github.com/google/mediapipe/issues/1889
[1] : https://github.com/bazelbuild/bazel/issues/12702


libraua:
There is lts_2021_03_24.2 already

PR #2101, opened 2021-05-31
Is functions are const funtctions in port.h

PeterPocsi:
All the "Is" functions should be const functions (like several other like IsDone).


PeterPocsi:
@googlebot I fixed it.

PeterPocsi:
@googlebot I fixed it.

jiuqiant:

jiuqiant:

PR #2154, opened 2021-06-09
fix pose code snippet

gabrielsanchez:
First issue: mp_holistic is not defined in this snippet. It was replaced by mp_pose.
Second issue: `model_complexity=2` is available for MediaPipe Holistic but not for MediaPipe Pose. 

The snippet will work with the changes that I made.


gabrielsanchez:
@googlebot I signed it!

jiuqiant:

PR #2292, opened 2021-07-15
fix installation guide

daniel13520cs:
use bazelisk would fix the build issue(bazel does not ignore .bazelversion file)


jiuqiant:

jiuqiant:

PR #2347, opened 2021-07-28
Update pose.md (Missing a DOM element for LandmarkGrid)

TomHsiao1260:
In JavaScript Solution API demo, seems that it missed a DOM element for LandmarkGrid class to initialize.


jiuqiant:

PR #2443, opened 2021-08-20
Set byte order of audio data.

H1Gdev:
In [AudioRecord#read](https://developer.android.com/reference/android/media/AudioRecord#read(java.nio.ByteBuffer,%20int))

`and will be native endian.`


mediapipe/java/com/google/mediapipe/components/MicrophoneHelper.java
```diff
@@ -181,6 +182,7 @@ private void setupAudioRecord() {
 
                 // TODO: Fix audio data cloning.
                 ByteBuffer audioData = ByteBuffer.allocateDirect(audioPacketBufferSize);
+                audioData.order(ByteOrder.nativeOrder());
```
H1Gdev, line 185: Users need to set byte order.

PR #2450, opened 2021-08-22
Fix anchor case

GantMan:
Anchor doesn't work on the website because anchors are case-sensitive.  This fixes that mistake for a link.


jiuqiant:

jiuqiant:

Issue #1965, opened 2021-05-01
ImageToTensor Error on GPU example for desktop.


txf-:
I'm running this on Manjaro Linux, Nvidia Driver version 460.73.1, Mediapipe version 0.8.3.2. The example executes then terminates with an error in the graph.

`Calculator::Process() for node "handlandmarktrackinggpu__handlandmarkgpu__ImageToTensorCalculator" failed: ; ROI width and height must be > 0.ulators/tensor/image_to_tensor_utils.cc:54) `

Here is the entire log output:

```
GLOG_logtostderr=1 bazel-bin/mediapipe/examples/desktop/hand_tracking/hand_tracking_gpu   --calculator_graph_config_file=mediapipe/graphs/hand_tracking/hand_tracking_desktop_live_gpu.pbtxt
I20210501 19:18:16.535331 112382 demo_run_graph_main_gpu.cc:52] Get calculator graph config contents: # MediaPipe graph that performs multi-hand tracking with TensorFlow Lite on GPU.
# Used in the examples in
# mediapipe/examples/android/src/java/com/mediapipe/apps/handtrackinggpu.

# GPU image. (GpuBuffer)
input_stream: "input_video"

# GPU image. (GpuBuffer)
output_stream: "output_video"
# Collection of detected/predicted hands, each represented as a list of
# landmarks. (std::vector<NormalizedLandmarkList>)
output_stream: "hand_landmarks"

# Generates side packet cotaining max number of hands to detect/track.
node {
  calculator: "ConstantSidePacketCalculator"
  output_side_packet: "PACKET:num_hands"
  node_options: {
    [type.googleapis.com/mediapipe.ConstantSidePacketCalculatorOptions]: {
      packet { int_value: 2 }
    }
  }
}

# Detects/tracks hand landmarks.
node {
  calculator: "HandLandmarkTrackingGpu"
  input_stream: "IMAGE:input_video"
  input_side_packet: "NUM_HANDS:num_hands"
  output_stream: "LANDMARKS:hand_landmarks"
  output_stream: "HANDEDNESS:handedness"
  output_stream: "PALM_DETECTIONS:palm_detections"
  output_stream: "HAND_ROIS_FROM_LANDMARKS:hand_rects_from_landmarks"
  output_stream: "HAND_ROIS_FROM_PALM_DETECTIONS:hand_rects_from_palm_detections"
}

# Subgraph that renders annotations and overlays them on top of the input
# images (see hand_renderer_gpu.pbtxt).
[...]

Issue #2251, opened 2021-07-04
Pose Tracking (GPU) and Holistic Tracking (GPU) crash on Linux


homuler:
**System information** (Please provide as much relevant information as possible)

- Have I written custom code (as opposed to using a stock example script provided in MediaPipe): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04, Android 11, iOS 14.4): Arch Linux (version 5.12.13-arch1-2 (gcc (GCC) 11.1.0, GNU ld (GNU Binutils) 2.36.1))
- Programming Language and version ( e.g. C++, Python, Java): C++ (gcc (Arch Linux 10.2.0-3) 10.2.0
, g++ (Arch Linux 10.2.0-3) 10.2.0)
- [MediaPipe version](https://github.com/google/mediapipe/releases): v0.8.6
- Bazel version (if compiling from source): 3.7.2 (downloaded)
- Solution ( e.g. FaceMesh, Pose, Holistic ): Pose, Holistic

**Describe the current behavior:**
Pose Tracking (GPU) and Holistic Tracking (GPU) examples crash.

**Describe the expected behavior:**
Pose Tracking (GPU) and Holistic Tracking (GPU) examples work on Linux.

**Standalone code to reproduce the issue:**
Build with the below command
```sh
bazel build -c opt --copt -DMESA_EGL_NO_X11_HEADERS --copt -DEGL_NO_X11 //mediapipe/examples/desktop/pose_tracking:pose_tracking_gpu
```

and run
```sh
GLOG_logtostderr=1 ./bazel-bin/mediapipe/examples/desktop/pose_tracking/pose_tracking_gpu --calculator_graph_config_file=mediapipe/graphs/pose_tracking/pose_tracking_gpu.pbtxt
```

**Other info / Complete Logs :**
```txt
I20210704 19:24:31.353058 139303 demo_run_graph_main_gpu.cc:52] Get calculator graph config contents: # Tracks and renders pose + hands + face landmarks.

# GPU buffer. (GpuBuffer)
input_stream: "input_video"

# GPU image with rendered results. (GpuBuffer)
output_stream: "output_video"

# Throttles the images flowing downstream for flow control. It passes through
# the very first incoming image unaltered, and waits for downstream nodes
# (calculators and subgraphs) in the graph to finish their tasks before it
# passes through another image. All images that come in while waiting are
# dropped, limiting the number of in-flight images in most part of the graph to
[...]

PR #2464, opened 2021-08-26
fix: check if ROI width and height is NaN

homuler:
Fixes https://github.com/google/mediapipe/issues/1965

When `//mediapipe/examples/desktop/hand_tracking:hand_tracking_gpu` or `//mediapipe/examples/desktop/pose_tracking:pose_tracking_gpu` is run on Linux, it crashes sometimes, especially just after camera starts or camera moves rapidly.
This issue occurs because widths and heights of detection's bounding boxes can be NaN.
This issue may be GPU-dependent, and I confirmed that it occurs with GeForce GTX TITAN X (driver version: 470.63.01).

Note that GPU sample (e.g. Pose Tracking, Holistic) may crash on Linux (cf. #2251) regardless of this.


jiuqiant:

brettkoonce:
ty @homuler!

jiuqiant:

ROBYER1:
Does this stuff ever get backported? We are using 0.8.4 at the moment and want to incorporate this patch

PR #2493, opened 2021-09-01
Added code to extract data(Image->CSV)

arpita739:
I have added a code that can extract landmarks of Hands from Images and create a CSV file to store it for training using Machine Learning algorithms.


arpita739:
@googlebot I signed it!

PR #2576, opened 2021-09-22
Update .bazelrc

aishwarya123mathpati:
To make MediaPipe get TensorFlow’s CUDA settings copied the content from tensorflow's bazel.rc file into mediapipe's  bazel.rc file


aishwarya123mathpati:
@googlebot I signed it!

PR #2608, opened 2021-09-28
Fixes #2489 Require either PBUFFER or WINDOW support in EGL, never both.

### LINKED ISSUE #2489: Support for EGL's drm/gbm platform for GPU inference on headless unix systems without a compositor
EGL from mesa (version 21) on our RPI 4s (OS is poky built with yocto) provides a "GBM platform" for rendering without a compositor like X or Wayland running. Rendering directly to displays without a compositor like this is simpler and faster for embedded applications.

However mediapipe currently doesn't work with the env var `EGL_PLATFORM=drm` as mesa's drm platform implementation doesn't support pbuffers. After browsing the source it seems like mediapipe doesn't use that pbuffer surface for anything except `eglMakeCurrent` which should be unnecessary (can use `EGL_NO_SURFACE` instead).

tflite also instantiates a pbuffer surface but I haven't yet dug deep enough to find out if it's neccessary or similarly pointless.

I'm opening this issue in-case anyone else comes across this problem with runtime errors like: `eglChooseConfig() returned no matching EGL configuration for RGBA8888 D16 ES3 request.`

By using `eglinfo` you can see in the `supported surfaces` section that mesa's GBM platform only supports `win` but x11 supports `win,pb,pix`

BTW nvidea's non-free drivers support `pb,str` in their device platform so switching to window surfaces instead of pbuffer surfaces would be non-portable.

willstott101:
I'm not 100% sure if pbuffers are used elsewhere specifically in the codebase but this has worked to allow me to use GBM. I'll test this shortly on a headless system without X installed at all.

Would `EglManager.java` be affected by this too? We have no desire to try using Java with mediapipe right now so I haven't tested it.


willstott101:
Further cleaning up of the gl_context headers and gl_context_egl specifically is possible thanks to this - but I've left the patch utterly minimal for now.

willstott101:
Keep in mind that if you have X or Wayland running you don't need to use this, but this patch fixes `EGL_PLATFORM=gbm`, which is only necessary on systems without displays, where running X or Wayland makes no sense.

willstott101:
Rebased and updated to remove the need for `MEDIAPIPE_OMIT_EGL_WINDOW_BIT` C flag. by trying multiple configs in a short for loop.

wndfly:
This modify is very useful, it can solve my problem that run in docker with errors: 
 `RuntimeError: ; eglChooseConfig() returned no matching EGL configuration for RGBA8888 D16 ES2 request.` 
`RuntimeError: ; Unable to initialize EGLpipe/gpu/gl_context_egl.cc:161)`.
 Thank you!

linyu0219:
Thank you, the same issue like above Juggernaut.

PR #2712, opened 2021-10-26
Fixed ifdef-else error for mac.

holohiko:


schmidt-sebastian:
This was due to a build configuration mismatch on M1. This has now been fixed.

PR #2743, opened 2021-11-05
Python face mesh module: Use tuples rather than sets in order to preserve connection ordering

trzy:
Face mesh connectivity is presently stored in sets, which are unordered and therefore make recovery of the mesh very difficult. It would be better just to use immutable tuples to convey this information.


trzy:
@googlebot I signed it!

trzy:
@googlebot I signed it!

PR #2748, opened 2021-11-06
fix add_packet_to_input_stream deadlock of python

Zondabb:
I found this python code will cause deadlock:
```
import mediapipe.python as mp

def assign_packet(stream_name, packet):
    """
    assign packet
    """
    output_str = mp.packet_getter.get_str(packet)
    print('get one packet: {}'.format(output_str))


if __name__ == "__main__":
    text_config = """
        input_stream: "in"
        output_stream: "out"
        node {
            calculator: "PassThroughCalculator"
            input_stream: "in"
            output_stream: "out"
        }
    """
    hello_world_packet = mp.packet_creator.create_string('hello world!')
    graph = mp.CalculatorGraph(graph_config=text_config)
    graph.observe_output_stream('out', assign_packet)
    graph.start_run()

    for i in range(1000):
        graph.add_packet_to_input_stream(stream='in', packet=hello_world_packet, timestamp=i)
        print('put one packet')
    graph.wait_until_idle()
    print('done!')
```

It's because the inference `add_packet_to_input_stream` do not release the GIL of python. It will block when the input queue is full.  The output callback function can't  acquire the GIL and block.
I fix this bug by release the GIL in `add_packet_to_input_stream`. I also add the same code in `wait_for_observed_output` for multi-threaded situation.

### DIFF
--- a/mediapipe/python/pybind/calculator_graph.cc
+++ b/mediapipe/python/pybind/calculator_graph.cc
@@ -168,6 +168,7 @@ void CalculatorGraphSubmodule(pybind11::module* module) {
         RaisePyErrorIfNotOk(
             self->AddPacketToInputStream(stream, packet.At(packet_timestamp)));
       },
+      py::call_guard<py::gil_scoped_release>(),
       R"doc(Add a packet to a graph input stream.
 
   If the graph input stream add mode is ADD_IF_NOT_FULL, the packet will not be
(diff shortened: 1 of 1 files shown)


Zondabb:
@googlebot I signed it!

Zondabb:
@googlebot I fixed it.

Zondabb:
@googlebot I fixed it.

jiuqiant:
Thanks for reporting this. We also fixed it internally and it's released in https://github.com/google/mediapipe/commit/e6c19885c6d3c6f410c730952aeed2852790d306.

PR #2941, opened 2021-12-31
[change] RET_CHECK_EQ -> RET_CHECK_GE

lilinxiong:
The number of landmarks used in face_geometry is not less than the number it needs, which is 468. Then create a matrix according to canonical_metric_landmarks_.cols(). Even if the subsequent [geometry_pipeline_metadata_landmarks.pbtxt](https://github.com/google/mediapipe/blob/v0.8.8/mediapipe/modules/face_geometry/data/geometry_pipeline_metadata_landmarks.pbtxt) is modified (decreased or increased), it will not affect the normal process.


lilinxiong:
I signed the CLA

PR #2944, opened 2021-12-31
[face_geometry] [change] RET_CHECK_EQ -> RET_CHECK_GE

lilinxiong:
The number of landmarks used in face_geometry is not less than the number it needs, which is 468. Then create a matrix according to canonical_metric_landmarks_.cols(). Even if the subsequent [geometry_pipeline_metadata_landmarks.pbtxt](https://github.com/google/mediapipe/blob/v0.8.8/mediapipe/modules/face_geometry/data/geometry_pipeline_metadata_landmarks.pbtxt) is modified (decreased or increased), it will not affect the normal process.


wingdi:
Hi, Thanks for your commit.
When i add vertex normal in 3d effect model(.Mesh3D) / geometry_pipeline_metadata_landmarks.binarypb / geometry_pipeline_metadata_detection.binarypb,  and use this commit, it didn't show camera's frame, only show effect model.  Using main branch also have this problem.
 
151645337878_ pic

PR #3316, opened 2022-05-03
Use absl::Status and co explicitly

kaliaumem:
MediaPipe exposes `mediapipe::Status`, `mediapipe::StatusCode`, `mediapipe::StatusOr<>` and `mediapipe::OkStatus()` which are aliases to their Abseil counterparts. Those aliases have been marked as deprecated, as shown here, for example: https://github.com/google/mediapipe/blob/c6c80c37452d0938b1577bd1ad44ad096ca918e0/mediapipe/framework/deps/status.h#L30

When building with warnings turned on, deprecation warnings end up spamming the build logs quite a lot, and play a part in being unable to treat warnings as errors in external code that includes some of MediaPipe's public headers.

This PR aims at removing all the uses I could find internally in the MediaPipe codebase.
Ideally I'd also like to remove the aliases themselves, but I realize this could break 3rd party code despite the deprecation flag because the warnings are silenced by default in `.bazelrc`, so I left them in.
I also left https://github.com/google/mediapipe/blob/master/mediapipe/framework/deps/status_test.cc mostly untouched, although 95% of it only tests Abseil's implementation. The test cases I did not touch could probably be safely deleted.

### DIFF
--- a/mediapipe/calculators/core/matrix_to_vector_calculator.cc
+++ b/mediapipe/calculators/core/matrix_to_vector_calculator.cc
@@ -58,7 +58,7 @@ MEDIAPIPE_REGISTER_NODE(MatrixToVectorCalculator);
 
 absl::Status MatrixToVectorCalculator::Open(CalculatorContext* cc) {
   cc->SetOffset(0);
-  return mediapipe::OkStatus();
+  return absl::OkStatus();
 }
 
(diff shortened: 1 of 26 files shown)


TimoPtr:
@jiuqiant @chuoling @sgowroji Would it be possible to merge this?

PR #3613, opened 2022-08-16
[fix] sobel_edges_calculator.cc

AstroWYH:


AstroWYH:
@sgowroji @chuoling @jiuqiant

PR #3893, opened 2022-11-30
[NOT FOR MERGING] iOS Task Library Prototype

priankakariatyml:
1. Prototyped the iOS task library by developing the text classifier APIs.
2. Only includes text classifier init methods.

### DIFF
--- a/mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
+++ b/mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
@@ -1,0 +1,44 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================*/
+#import <Foundation/Foundation.h>
+
+#include "mediapipe/framework/calculator.pb.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * The base class of the user-facing iOS mediapipe text task api classes.
+ */
+NS_SWIFT_NAME(BaseTextTaskApi)
+@interface MPPBaseTextTaskApi : NSObject
+
+/**
+ * Initializes a new `MPPBaseTextTaskApi` with the mediapipe text task graph config proto.
+ *
+ * @param graphConfig A mediapipe text task graph config proto.
+ *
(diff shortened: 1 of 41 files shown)


priankakariatyml:
@schmidt-sebastian @khanhlvg Raised PR for iOS task library prototype

mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
```diff
@@ -0,0 +1,48 @@
...
+#include "mediapipe/framework/calculator.pb.h"
+#include "mediapipe/tasks/cc/core/task_runner.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * The base class of the user-facing iOS mediapipe text task api classes.
+ */
+NS_SWIFT_NAME(BaseTextTaskApi)
+@interface MPPBaseTextTaskApi : NSObject {
```
priankakariatyml, line 26: @schmidt-sebastian @khanhlvg 
This is the base class for all text APIs. Please note that C++ header files are included here.  These headers and any headers transitively imported by them will have to be part of the iOS framework we build for distribution. According to me this is not a good design as consumers of the API will be exposed to many C++ headers. Also I am yet to check the feasibility of this design when building CocoaPods. 

**My Suggestion:**
 Instead of extending the individual tasks from a  base task api, we can create a base task util class which will encapsulate the c++ task runner and all the common functionality needed. The utils will only be included in the .mm file. So we don't have to expose the C++ headers.

mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
```diff
@@ -0,0 +1,48 @@
...
+ @protected
+  std::unique_ptr<mediapipe::tasks::core::TaskRunner> cppTaskRunner;
```
priankakariatyml, line 28: @schmidt-sebastian @khanhlvg @lu-wang-g
I am using the C++ task runner like the Python implementation. Java on the other hand, creates its own task runner with the graph APIs. 
If need be, we can follow the Java approach. Only thing is, we will have to rewrite  [MPPGraph](https://github.com/google/mediapipe/blob/master/mediapipe/objc/MPPGraph.h#L25) and add a lot of other classes to process packets etc. The current iOS graph APIs depend on `MPPGraph`. If we use this approach, we will have to ensure, we don't break anything.

LMK if sticking to the Python approach is fine.

schmidt-sebastian (approved):
Approving this so I can test importing. This is not yet ready to be merged.

mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
schmidt-sebastian: I think we will be much better off in the long run if we follow the Python approach and re-use as much code as possible.

mediapipe/tasks/ios/text/core/sources/MPPBaseTextTaskApi.h
schmidt-sebastian: Can you elaborate on this? The headers will be part of the build process, but does the user have to consume them in order to use our iOS API?

PR #3973, opened 2023-01-06
Added iOS Classification Result, Category and Helper Classes

priankakariatyml:
1. Added MPPClassificationResult, MPPCategory and their helpers
2. Removed MPPClassifierOptions and helpers since we have decided to in-line them in text classifier options to maintain consistency with Java and Python


schmidt-sebastian (approved):

mediapipe/tasks/ios/components/containers/BUILD
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 The MediaPipe Authors. All Rights Reserved.
+# Copyright 2023 The MediaPipe Authors. All Rights Reserved.
```
schmidt-sebastian, line 1: FYI We should keep the copyright year for already existing files.

schmidt-sebastian: Sorry I just saw that this is a branched file. Please disregard.

PR #4027, opened 2023-01-24
Fix the `cosine_similarity` utility function to work with quantized embeddings in the Python Tasks API 

kinaryml:
- Currently the cosine_similarity utility function doesn't work with quantized embeddings (generates 0)

Fix: Cast the int8 embeddings to float before computing cosine similarity on the embeddings.


khanhlvg:
@jiuqiant  PTAL when you're back to office.

@schmidt-sebastian FYI

jiuqiant (approved):

PR #4030, opened 2023-01-25
Added iOS Text Embedder Result and Options

priankakariatyml:
1. Added MPPEmbedding, MPPEmbeddingResult, MPPTextEmbedderResult
2. Added MPPTextEmbedderOptions


priankakariatyml:
@khanhlvg @schmidt-sebastian Raised a PR for adding embedding containers and options

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.h
```diff
@@ -0,0 +1,70 @@
...
+ * @brief The Floating-point embedding.
+ * Empty if the embedder was configured to perform scalar quantization.
+ */
+@property(nonatomic, readonly, nullable) float *floatEmbedding;
+
+/**
+ * @brief The Quantized embedding.
+ * Empty if the embedder was not configured to perform scalar quantization.
+ */
+@property(nonatomic, readonly, nullable) char *quantizedEmbedding;
```
priankakariatyml, line 38: @schmidt-sebastian @khanhlvg `quantizedEmbedding` is of type bytes in the proto. It is converted to `std::string` in C++. I have not used NSString* because if any operations are to be performed with the embedding, then the consumers may have to convert it into a char* array. Instead I am using char * directly as the type here. Moreover the embedding may be long since this is intended to use with image embedder as well. In case of streaming operations, continuous conversion to NSString may present extra over head.

mediapipe/tasks/ios/components/containers/sources/MPPEmbeddingResult.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * An Array of `MPPEmbedding` objects containing the embedding results for each head of the model.
+ */
+@property(nonatomic, readonly) NSArray<MPPEmbedding *> *embeddings;
+
+/**
+ * @brief The optional timestamp (in milliseconds) of the start of the chunk of data corresponding
+ * to these results.
+ * This is only used for embedding extraction on time series (e.g. audio embedder). In these use
+ * cases, the amount of data to process might exceed the maximum size that the model can process: to
```
schmidt-sebastian, line 33: Can you rewrite this to remove the `:`?

mediapipe/tasks/ios/text/text_embedder/sources/MPPTextEmbedderOptions.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+NS_SWIFT_NAME(TextEmbedderptions)
+@interface MPPTextEmbedderOptions : MPPTaskOptions <NSCopying>
+
+/**
+ * @brief Sets whether L2 normalization should be performed on the returned embeddings.
+ * Use this option only if the model does not already contain a native L2_NORMALIZATION TF Lite Op.
+ * In most cases, this is already the case and L2 norm is thus achieved through TF Lite inference.
+ *
+ * NO by default.
```
schmidt-sebastian, line 32: Should NO be in backticks? (Same below)

schmidt-sebastian (commented):
Thank you!

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.h
priankakariatyml: `NSData` is a wrapper class for any byte buffer.  It is not a primitive type unlike a `byte`  array in java. The primitive equivalent is `char *` like in C. 
Advantage of `NSData` is that for iOS devs it is a familiar type. Types like `char` `char *` etc. are not used in day to day tasks. 
Plus if initialised in the right way, it will take care of deallocating the underlying byte buffer. But In our case,  even if we choose `char *,` the task library will be handling deallocation of this property once `MPPEmbedding` goes out of scope. So that's not really a problem.
Disadvantage is that`NSData` in our case is an overhead for performance. Especially since we can expect embeddings to be used for some other task.  For this we will have to directly access and manipulate the underlying byte buffer.  So converting the `std::string` to `NSData` only for the consumers of the API to access the pointer seems costly. Also devs dealing with unconventional tasks like manipulation of embeddings will be expected to handle these kind of types and their deallocation if, they choose to copy. 

So the question is whether we should use `NSData` for the sake of familiarity?

(https://developer.apple.com/documentation/foundation/nsdata?language=objc)
I noticed that web is using a [`uint8 array`
](https://github.com/google/mediapipe/blob/master/mediapipe/tasks/web/components/containers/embedding_result.d.ts#L35). We can use int8* instead of char * if need be.

priankakariatyml: `NSData` is a wrapper class for any byte buffer.  It is not a primitive type. Unlike `byte` array in java. Primitive equivalent is `char *` like in C. 

**Advantages of `NSData`**

- Familiar type for iOS devs. Types like `char *` are not used in regular tasks. 
- If initialised in the right way, it will take care of deallocating the underlying byte buffer. If we choose `char *,` the task library will be handling deallocation of this property once `MPPEmbedding` goes out of scope. So that's not really a problem.

**Disadvantage**
Overhead for performance. Especially since we can expect embeddings to be used for some other task.  For this we will have to directly access and manipulate the underlying byte buffer.  So converting the `std::string` to `NSData` only for the consumers of the API to access the pointer seems costly. 

Devs dealing with unconventional tasks like integrating embedding models will be expected to handle these kind of types and their deallocation if, they choose to copy. 
So the question is whether we should use `NSData` for the sake of familiarity?

I noticed that web is using a [`uint8 array`
](https://github.com/google/mediapipe/blob/master/mediapipe/tasks/web/components/containers/embedding_result.d.ts#L35). We can use `int8*` instead of `char *` since that makes more sense for quantized values.

khanhlvg: In the TFLite Task Lib image segmenter API, we used float array pointer directly to represent the segmentation masks only because it's more performant than using high-level data type like NSData or NSArray, and the segmentation masks are large enough to make those performance difference noticeable. However, embedding vector are usually small and I don't think there's a need to proceed embedding values independently like in segmentation mask, so using `NSData` sounds like a good idea to me.

schmidt-sebastian: It looks like we are creating a copy of the mask even in the ImageSegmenter (tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m). This avoids the lifetime issue and is a good choice here since the masks are float values. In the embedding case, since we are talking about bytes, I feel that we should just use NSData (and copy). This avoids lifetime issues and makes our API much more usable (otherwise, we would have to return a pointer and the length).

priankakariatyml: I have updated `quantizedEmbedding` to be an `NSData`. Considering the above feedback, `floatEmbedding` has been changed to `NSArray<NSNumber *>*`. Note that to keep these types uniform, we can make them both `NSArray`s or `NSData`s if need be. LMK your thoughts.

schmidt-sebastian (approved):

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param floatEmbedding The optional Floating-point embedding.
+ * @param quantizedEmbedding The optional Quantized embedding.
+ * @param headIndex The index of the embedder head.
+ * @param headName The optional name of the embedder head.
+ *
+ * @return An instance of `MPPEmbedding` initialized with the given float embedding, quantized
+ * embedding, head index and head name.
+ */
+- (instancetype)initWithFloatEmbedding:(nullable NSArray<NSNumber *> *)floatEmbedding
```
khanhlvg, line 58: @schmidt-sebastian @priankakariatyml sorry that I missed this one from the beginning. Given that we use NSData to represent quantizedEmbedding, should floatEmbedding be NSData as well for consistency? 

If floatEmbedding should be `NSArray<NSNumber>` for consistency with C++ layer, we can use NSArray<NSNumber *> for quantizedEmbedding as well because NSNumber also supports uint8.

Besides, we can have readonly property that converts the

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.m
```diff
@@ -0,0 +1,33 @@
...
+
+@implementation MPPEmbedding
+
+- (instancetype)initWithFloatEmbedding:(nullable NSArray<NSNumber *> *)floatEmbedding
+                    quantizedEmbedding:(nullable NSData *)quantizedEmbedding
+                             headIndex:(NSInteger)headIndex
+                              headName:(nullable NSString *)headName {
+  self = [super init];
+  if (self) {
+    _headIndex = headIndex;
```
khanhlvg, line 25: @priankakariatyml @schmidt-sebastian As an embedding can't be both quantized and float, should we do a check here to make sure that MPPEmbedding isn't initialized with both float and quantized embedding?

mediapipe/tasks/ios/text/text_embedder/sources/MPPTextEmbedderOptions.h
```diff
@@ -0,0 +1,47 @@
...
+#import <Foundation/Foundation.h>
+
+#import "mediapipe/tasks/ios/core/sources/MPPTaskOptions.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Options for setting up a `MPPTextEmbedder`.
+ */
+NS_SWIFT_NAME(TextEmbedderptions)
```
khanhlvg, line 24: typo: TextEmbedderOptions

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.h
schmidt-sebastian: I think that NSData is not the right choice for float numbers (since we actually want to treat them as floats to compute the cosine similarity), but NSData feels right to me for the Uint8 (byte) data. Since both you and Prianka seem to prefer NSArray both, I will update in the import CL.

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.m
schmidt-sebastian: Prianka asked me in an earlier comment. I think it is okay not to do so.

mediapipe/tasks/ios/text/text_embedder/sources/MPPTextEmbedderOptions.h
schmidt-sebastian: Will fix.

mediapipe/tasks/ios/components/containers/sources/MPPEmbedding.h
priankakariatyml: Besides if we represent float numbers as `NSData`, we will need an additional property for returning the length of the embedding. `NSData` only gives out length in bytes.

PR #4066, opened 2023-02-06
Added MPPTaskImage for iOS Vision Tasks

priankakariatyml:


priankakariatyml:
@khanhlvg @schmidt-sebastian 
I have raised a PR for MPPTaskImage

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
```diff
@@ -0,0 +1,178 @@
...
+@property(nonatomic, readonly) CGFloat height;
+
+/**
+ * The display orientation of the image. If `imageSourceType` is `MPPTaskImageSourceTypeImage`, the
+ * default value is `image.imageOrientation`; otherwise the default value is `UIImageOrientationUp`.
+ * If the `MPPTaskImage` is being used as input for any MediaPipe vision tasks and is set to any
+ * orientation other than `UIImageOrientationUp`, inference will be performed on a rotated copy of
+ * the image according to the orientation.
+ */
+@property(nonatomic, readonly) UIImageOrientation orientation;
```
priankakariatyml, line 47: @khanhlvg  @schmidt-sebastian [`MLImage`](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/odml/ios/image/apis/GMLImage.h#L44)'s orientation can be set any time. 
I have made this property readonly and added designated initializers with provision to pass in orientation only at the time of initialisation. iOS classes like [UIImage](https://developer.apple.com/documentation/uikit/uiimage?language=objc) and [VNImageRequestHandler](https://developer.apple.com/documentation/vision/vnimagerequesthandler?language=objc) implement orientation this way.  
This will avoid any issues arising from changing the image orientation during the life time of an image. Say, after displaying the image in a certain orientation, it is set to a different value before using it for inference. Results may be different from what you see on screen for certain tasks if the developer forgets to handle this.

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
```diff
@@ -0,0 +1,178 @@
...
+ * @return A new `MPPTaskImage` instance with the given pixel buffer as the source. `nil` if the
+ * given pixel buffer is `nil` or invalid.
+ */
+- (nullable instancetype)initWithPixelBuffer:(CVPixelBufferRef)pixelBuffer
+                                       error:(NSError **)error NS_DESIGNATED_INITIALIZER;
+
+/**
+ * Initializes an `MPPTaskImage` object with the given pixel buffer and orientation.
+ *
+ * If the newly created `MPPTaskImage` is used as input for any MediaPipe vision tasks, inference
```
priankakariatyml, line 119: @schmidt-sebastian @khanhlvg  I am hoping my understanding of the necessity of this property is correct. Please have a look.

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
priankakariatyml: @khanhlvg @schmidt-sebastian Java has [ImageProcessing](https://github.com/google/mediapipe/blob/master/mediapipe/tasks/java/com/google/mediapipe/tasks/vision/core/ImageProcessingOptions.java#L87) with `rotationDegrees`. This seems to serve the purpose of `orientation`, if my understanding of its usage is correct. 
We must omit this class from the iOS implementation since `roi` is the only other property which can be directly passed in. LMK your thoughts.

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
```diff
@@ -0,0 +1,178 @@
...
+/** Image source is a `UIImage`. */
+static const MPPTaskImageSourceType MPPTaskImageSourceTypeImage = 0;
+/** Image source is a `CVPixelBuffer`. */
+static const MPPTaskImageSourceType MPPTaskImageSourceTypePixelBuffer = 1;
+/** Image source is a `CMSampleBuffer`. */
+static const MPPTaskImageSourceType MPPTaskImageSourceTypeSampleBuffer = 2;
+
+/** An image used in on-device machine learning using MediaPipe Task library. */
+NS_SWIFT_NAME(MPPTaskImage)
+@interface MPPTaskImage : NSObject
```
khanhlvg, line 32: Please name it MPPImage for Obj-C and MPImage for Swift (to avoid collision with SwiftUI's [Image](https://developer.apple.com/documentation/swiftui/image))

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
schmidt-sebastian: The image tasks only support 90, 180, 270 as orientation. Using UIImageOrientation seems sensible. We might need to error checking for [upMirrored](https://developer.apple.com/documentation/uikit/uiimage/orientation/upmirrored) and the related properties.

priankakariatyml: @schmidt-sebastian Should we check it here, or in a util class that feeds this image to the vision tasks, considering this will not be used only as input to vision tasks.

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
priankakariatyml: Done

mediapipe/tasks/ios/vision/core/sources/MPPTaskImage.h
schmidt-sebastian: I would probably check in the Vision tasks. We also need to determine how we can support "Region Of Interest".

schmidt-sebastian (approved):

PR #4075, opened 2023-02-08
Added Objective C tests for iOS ML Image

priankakariatyml:


mediapipe/tasks/ios/test/vision/core/MPPImageTests.m
```diff
@@ -0,0 +1,328 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+static NSString *const kTestImageName = @"burger";
+static NSString *const kTestImageType = @"jpg";
+static CGFloat kTestImageWidthInPixels = 480.0f;
+static CGFloat kTestImageHeightInPixels = 325.0f;
+static NSString *const kExpectedErrorDomain = @"com.google.mediapipe.tasks";
+
+#define AssertEqualErrors(error, expectedError)                                               \
```
schmidt-sebastian, line 32: FYI (for future PRs): The last internal reviewer asked us to create functions and not rely on macros so much:

"Avoid macros, especially where const variables, enums, Xcode snippets, or C functions may be used instead.

Macros make the code you see different from the code the compiler sees. Modern C renders traditional uses of macros for constants and utility functions unnecessary. Macros should only be used when there is no other solution available."

Macros are only really supposed to be used as compile-time guards.

No need to fix for any current PRs/current code (though a cleanup at a later point would be nice).

mediapipe/tasks/ios/test/vision/core/MPPImageTests.m
```diff
@@ -0,0 +1,328 @@
...
+  MPPImage *mppImage = [[MPPImage alloc] initWithUIImage:nil error:&error];
+  [self
+      assertInitFailsWithImage:mppImage
+                         error:error
+                 expectedError:[NSError errorWithDomain:kExpectedErrorDomain
+                                                   code:MPPTasksErrorCodeInvalidArgumentError
+                                               userInfo:@{
+                                                 NSLocalizedDescriptionKey : @"Image cannot be nil."
+                                               }]];
+  NSLog(@"Errorrrr........%@", error.localizedDescription);
```
schmidt-sebastian, line 117: Please remove

schmidt-sebastian (approved):

PR #4211, opened 2023-03-28
Added iOS Image Classifier Swift Tests

priankakariat:


mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
```diff
@@ -0,0 +1,690 @@
...
+
+    try assertResultsForClassifyImage(
+      mpImage,
+      usingImageClassifier: imageClassifier,
+      hasCategoryCount: expectedCategoryCount,
+      andCategories: expectedCategories
+    )
+  }
+
+  func testCreateImageClassifierWithCategoryAllowlistAndDenylistFails() throws {
```
schmidt-sebastian, line 217: This seems to be missing `testCreateImageClassifierWithMissingModelPathFails`

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
```diff
@@ -0,0 +1,690 @@
...
+          timestampMs: i))
+      try assertImageClassifierResult(
+        imageClassifierResult,
+        hasCategoryCount: maxResults,
+        andCategories: 
+          ImageClassifierTests.expectedResultsClassifyBurgerImageWithFloatModel
+      )
+    }
+  }
+}
```
schmidt-sebastian, line 690: Can we also port `testClassifyWithQuantizedModelSucceeds`?

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
priankakariatyml: That scenario will never happen in Swift and hence can't be tested. `modelPath` param is strictly `nonnull` and hence optionals cannot be passed to it.  Since Swift is strongly typed, If we try to instantiate a non optional path which doesn't point to any existing file using the native apis, an error is raised. 
Objective C only raises a warning if you pass nil to non null param. Hence this test was needed.

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
priankakariatyml: Added

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
```diff
@@ -687,4 +689,125 @@ class ImageClassifierTests: XCTestCase {
       )
     }
   }
+
+  func testClassifyWithOutOfOrderTimestampsAndLiveStreamModeSucceeds() throws {
```
schmidt-sebastian, line 721: s/succeeds/fails?

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
```diff
@@ -687,4 +689,125 @@ class ImageClassifierTests: XCTestCase {
...
+   // `expectation.expectedFulfillmentCount` times.
+   // If `expectation.isInverted = true`, the test will only succeed if 
+   // expectation is not fullfilled for the specified `expectedFulfillmentCount`.
+   // Since in our case we cannot predict how many times the expectation is 
+   // supposed to be fullfilled setting,
+   // `expectation.expectedFulfillmentCount` = `iterationCount` and 
+   // `expectation.isInverted = true` ensures that test succeeds if 
+   // expectation is not fullfilled `iterationCount` times.
+    let expectation = expectation(description: "liveStreamClassify")
+    expectation.expectedFulfillmentCount = iterationCount;
```
schmidt-sebastian, line 805: If I am reading this correctly this test would fail if there happens to be no flow limiting. The C++ test allows <= iterations, versus this only allows <iterations. Can we fix this?

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
```diff
@@ -687,4 +689,125 @@ class ImageClassifierTests: XCTestCase {
...
+      imageClassifierOptions))
+
+     let mpImage = try XCTUnwrap(
+      imageWithFileInfo(ImageClassifierTests.burgerImage))
+    
+    for i in 0..<iterationCount {
+      XCTAssertNoThrow(
+        try imageClassifier.classifyAsync(
+          image: mpImage,
+          timestampMs: i))
```
schmidt-sebastian, line 836: Can we work around flow limiting by passing a higher gap between timestamps (i.e. `i * 1000`)?

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
priankakariatyml: If it has no flow limiting, it will fail, since this test relies on the expectation not being fulfilled for the given iteration count. But we don't give the user an option to turn off flow limiting in live stream mode. In all the platforms this is the behaviour. So this test will not fail.
We can't enforce <= iterations with expectations in iOS and this is the only way to test async calls. I am still trying to figure out if there is an alternative to test async calls.

mediapipe/tasks/ios/test/vision/image_classifier/ImageClassifierTests.swift
schmidt-sebastian: I don't think the goal of this test should be to test flow limiting. We should test the classifyAsync call in LiveStream mode with multiple packages. The fact that there is flow limiting is an implementation detail (albeit one that makes the test harder to write). The current Swift test would fail if for some reason flow limiting was turned off or did no kick in, versus the C++ test would still succeed. I think we should try to figure out if we can achieve the same behavior on iOS (and maybe that's just raising the expectation limit to 101).

priankakariatyml: I have set it to 101 for now.

PR #4241, opened 2023-04-05
Interactive Segmenter MediaPipe Tasks Python  API

kinaryml:


jiuqiant (approved):
I will import this as is and ask the graph author to take a close look.

PR #4254, opened 2023-04-06
Added iOS Gesture Recognizer Result and Options

priankakariatyml:
1. Added MPPLandmark, MPPNormalizedLandmark and Helpers
2. Added MPPClassifierOptions and Helpers
3. Added MPPGestureRecognizer Options and Result


mediapipe/tasks/ios/components/containers/sources/MPPLandmark.m
```diff
@@ -0,0 +1,105 @@
...
+    _x = x;
+    _y = y;
+    _z = z;
+    _visibility = visibility;
+    _presence = presence;
+  }
+  return self;
+}
+
+- (NSUInteger)hash {
```
priankakariatyml, line 37: @schmidt-sebastian I have implemented hash similar to the Java implementation. Please note that we are converting the float into an `NSNumber` during computation of hash as apple does not provide a hash function for basic types like `float`.  If we don't perform this conversion, we will have to perform the underlying hashing logic of `NSNumber` which would be [something like this](https://stackoverflow.com/a/26433160).
since `isEqual` compares with a tolerance, we can't guarantee that for equal objects hash will be the same. Docs mention that [hash values of two equal objects should be same if it is intended to be used with hash tables](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418859-hash). 
But I guess even Java, Python etc., have the same requirement. So I am assuming this is okay.

schmidt-sebastian (approved):

PR #4268, opened 2023-04-10
Added iOS Object Detector Objective C Tests

priankakariatyml:
1. Removed Roi apis as object detection does not support rot
2. Fixed bug in check for whether roi is allowed in MPPVIsionTaskRunner.
3. Added Object Detector Objective C Tests

### DIFF
--- a/mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
+++ b/mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
@@ -421,12 +421,9 @@ static const float scoreDifferenceTolerance = 1e-2f;
 
   NSArray<MPPDetection *> *detections = @[
     [[MPPDetection alloc] initWithCategories:@[
-      [[MPPCategory alloc] initWithIndex:-1
-                                   score:0.750000f
-                            categoryName:@"teddy bear"
-                             displayName:nil],
+      [[MPPCategory alloc] initWithIndex:-1 score:0.699219f categoryName:@"cat" displayName:nil],
     ]
-                                 boundingBox:CGRectMake(0, 372, 416, 276)
+                                 boundingBox:CGRectMake(0, 608, 439, 387)
                                    keypoints:nil],
   ];
 
--- a/mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
+++ b/mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
@@ -1,0 +1,703 @@
+// Copyright 2023 The MediaPipe Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
```diff
@@ -0,0 +1,703 @@
...
+#import "mediapipe/tasks/ios/test/vision/utils/sources/MPPImage+TestUtils.h"
+#import "mediapipe/tasks/ios/vision/object_detector/sources/MPPObjectDetector.h"
+
+static NSString *const kModelName = @"coco_ssd_mobilenet_v1_1.0_quant_2018_06_29";
+static NSDictionary *const kCatsAndDogsImage = @{@"name" : @"cats_and_dogs", @"type" : @"jpg"};
+static NSDictionary *const kCatsAndDogsRotatedImage =
+    @{@"name" : @"cats_and_dogs_rotated", @"type" : @"jpg"};
+static NSString *const kExpectedErrorDomain = @"com.google.mediapipe.tasks";
+
+#define PixelDifferenceTolerance 5.0f
```
schmidt-sebastian, line 27: Let's use proper `const` variables for these.

mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
```diff
@@ -0,0 +1,703 @@
...
+#define ScoreDifferenceTolerance 1e-2f
+
+#define AssertEqualErrors(error, expectedError)                                               \
```
schmidt-sebastian, line 30: I think these macros are getting a bit more complicated than what I would think is acceptable for macros. Can we turn this into helper functions?

mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
```diff
@@ -0,0 +1,703 @@
...
+  MPPObjectDetectorOptions *options = [self objectDetectorOptionsWithModelName:kModelName];
+  options.maxResults = 1;
+
+  MPPObjectDetector *objectDetector = [self objectDetectorWithOptionsSucceeds:options];
+
+  NSArray<MPPDetection *> *detections = @[
+    [[MPPDetection alloc] initWithCategories:@[
+      [[MPPCategory alloc] initWithIndex:-1
+                                   score:0.750000f
+                            categoryName:@"teddy bear"
```
schmidt-sebastian, line 422: There is a bug in how we handle image rotation in all of our tasks. We are going through the APIs and fixing this. Can you see if you can port https://github.com/google/mediapipe/commit/e3185e3df05dc0ef347118c18e08b8a9ff5263d4 and verify that we are getting a "cat" here?

mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
priankakariatyml: Fixed

mediapipe/tasks/ios/test/vision/object_detector/MPPObjectDetectorTests.m
priankakariatyml: I have converted them into helper functions which use macros only for comparison of two categories and bounding boxes.

PR #4269, opened 2023-04-10
Migrate stale-master probot to Github action

shmishra99:
I've updated the workflow file to replace 'stale-master' probot. It will add 'stale' label to inactive issues/PRs if contains 'stat:awaiting response' label in case of further inactivity it will close the issue with proper comment.


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,66 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: 'actions/stale@v7'
+      with:
+        # Comma separated list of labels that can be assigned to issues to exclude them from being marked as stale.
+        exempt-issue-labels: 'override-stale'
+        # Comma separated list of labels that can be assigned to PRs to exclude them from being marked as stale.
+        exempt-pr-labels: "override-stale"
+        # Limit the No. of API calls in one run default value is 30.
+        operations-per-run: 500
```
ayushgdev, line 40: Why has the default limit been increased to 500?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,66 @@
...
+        # Prevent to remove stale label when PRs or issues are updated.
+        remove-stale-when-updated: false
```
ayushgdev, line 42: What is meant by "updated"? Does it only cover the comments made by issue author or any update like change of labels, comments, etc. by others as well? Because if it only covers original author, this should be true. Otherwise false.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,66 @@
...
+        # comment on issue if not active for more then 7 days.
+        stale-issue-message: 'This issue has been marked stale because it has no recent activity since 7 days. It will be closed if no further activity occurs. Thank you.'
+        # comment on PR if not active for more then 14 days.
+        stale-pr-message: 'This PR has been marked stale because it has no recent activity since 14 days. It will be closed if no further activity occurs. Thank you.'
+        # comment on issue if stale for more then 7 days.
+        close-issue-message: This issue was closed due to lack of activity after being marked stale for past 7 days.
+        # comment on PR if stale for more then 14 days.
+        close-pr-message: This PR was closed due to lack of activity after being marked stale for past 14 days.
+        # Number of days of inactivity before an Issue Request becomes stale
+        days-before-issue-stale: 7
```
ayushgdev, line 52: Can this be used in the stale-issue-message as a variable?

.github/workflows/stale.yaml
shmishra99: It will pick 3-4 issues/PRs only with default value 30 to stale/close. Increasing it to 500 so it will iterate over all the issues for stale/close.We have per hour API limit 5000/per workflow.

.github/workflows/stale.yaml
shmishra99: Any event will consider as a "Update" and workflow will remove the "stale" label.That's why keeping it as a false. 
For author comment remove stale we have probot running.

.github/workflows/stale.yaml
shmishra99: As per documentation we don't have such option available to declare a variable in stale/action.
Additionally all the stale messages are different. Even if we will create the variable for messages we need to create one variable for each messages.

ayushgdev:
@schmidt-sebastian Please review this PR.

schmidt-sebastian (approved):

kuaashish:
Hi @schmidt-sebastian,
Could you please merge this stale_management change request? Thank you

ayushgdev (approved):

PR #4280, opened 2023-04-13
Deprecated `output_type` and `Activation` for the Segmentation Python Tasks APIs

kinaryml:


mediapipe/tasks/python/test/vision/image_segmenter_test.py
```diff
@@ -127,20 +157,19 @@ def test_segment_succeeds_with_category_mask(self, model_file_type):
       raise ValueError('model_file_type is invalid.')
 
     options = _ImageSegmenterOptions(
-        base_options=base_options, output_type=_OutputType.CATEGORY_MASK)
+        base_options=base_options, output_category_mask=True)
```
yichunk, line 160: Please test with output_category_mask=True, output_confidence_masks=False.

mediapipe/tasks/python/test/vision/image_segmenter_test.py
```diff
@@ -280,20 +275,49 @@ def test_segment_for_video_with_out_of_order_timestamp(self):
...
 
+  def test_segment_for_video_in_confidence_mask_mode(self):
+    # Load the cat image.
+    test_image = _Image.create_from_file(
+      test_utils.get_test_data_path(
+        os.path.join(_TEST_DATA_DIR, _CAT_IMAGE)))
+
+    options = _ImageSegmenterOptions(
+      base_options=_BaseOptions(model_asset_path=self.model_path),
+      running_mode=_RUNNING_MODE.VIDEO)
```
yichunk, line 301: please explicitly set output_category_mask=False, output_confidence_masks=True.
and for category mode test, set output_category_mask=True, output_confidence_masks=False.
ditto for live stream mode.

yichunk (commented):
Thank Kinar, it looks good!
Leave some minor comments.

mediapipe/tasks/python/test/vision/image_segmenter_test.py
kinaryml: I happened to not set those flags because the C++ API happened to use the following defaults:

https://github.com/google/mediapipe/blob/78499b0bd0e4304bfb6eaa6029754681367673f9/mediapipe/tasks/cc/vision/image_segmenter/image_segmenter.h#L57
(The Python API uses the same defaults) 

But yes, I believe it's better to explicitly state the mode.

kinaryml:
@yichunk I've made the necessary changes now. PTAL at the commit and let know if it's all good now.

yichunk (approved):

yichunk:
Thank Kinar! approved.

mediapipe/tasks/python/vision/image_segmenter.py
```diff
@@ -64,27 +71,22 @@ class ImageSegmenterOptions:
...
       data. The result callback should only be specified when the running mode
       is set to the live stream mode.
   """
 
-  class OutputType(enum.Enum):
-    UNSPECIFIED = 0
-    CATEGORY_MASK = 1
-    CONFIDENCE_MASK = 2
-
   class Activation(enum.Enum):
```
yichunk, line 81: Activation setting is deprecated as well.

kinaryml: @yichunk I've removed it now.

yichunk (approved):
Thank Kinar!

PR #4291, opened 2023-04-14
Verify the results properly in the Object Detector Python Tasks API's tests

kinaryml:
Ensures tests for the Object Detector Python API to be in line with that of the [Java API](https://github.com/google/mediapipe/blob/89e6b824aefa019154caef829f876f2ba7fbc446/mediapipe/tasks/javatests/com/google/mediapipe/tasks/vision/objectdetector/ObjectDetectorTest.java#L56)

### DIFF
--- a/mediapipe/tasks/python/test/vision/object_detector_test.py
+++ b/mediapipe/tasks/python/test/vision/object_detector_test.py
@@ -113,6 +113,8 @@ _EXPECTED_DETECTION_RESULT = _DetectionResult(
         ),
     ]
 )
+_SCORE_DIFF_TOLERANCE = 1e-6
+_PIXEL_DIFF_TOLERANCE = 5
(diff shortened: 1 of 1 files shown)


mediapipe/tasks/python/test/vision/object_detector_test.py
```diff
@@ -136,6 +138,32 @@ def setUp(self):
         os.path.join(_TEST_DATA_DIR, _MODEL_FILE)
     )
 
+  def _assert_approximately_equals_bounding_box(self, bbox, expected_bbox):
+    self.assertEqual(bbox.width, expected_bbox.width)
+    self.assertEqual(bbox.height, expected_bbox.height)
```
kinaryml, line 143: Not sure if we need to account for any numerical instability errors in width and height here too. PTAL.

PR #4355, opened 2023-04-27
Added Building of iOS OpenCV XCFramework from sources to support M1 Macs

priankakariatyml:


third_party/opencv_ios_source.bzl
```diff
@@ -0,0 +1,150 @@
...
+    for (symlink_output, binary_file) in binaries_to_symlink:
+        ctx.actions.symlink(output = symlink_output, target_file = binary_file)
+        binary_symlink_files.append(symlink_output)
+    
+    # Return all the declared output files and symlinks as the output of this
+    # rule.
+    return [ DefaultInfo(files=depset(output_files + binary_symlink_files)) ]
+
+
+# This rule unzips an `opencv2.xcframework.zip` created by a genrule that
```
priankakariatyml, line 137: @schmidt-sebastian We can write custom rules which output directories. But `apple_static_xcframework_import` cannot consume directories. It can only take a list of files. Hence we must declare a full list of all files expected to be in opencv2.xcframework in this custom rule which will provide the headers to `apple_static_xcframework_import`. So hardcoding them cannot be avoided. Since we are fixing opencv at 4.5.1, this won't cause any problems.

schmidt-sebastian (approved):

priankakariatyml:
@schmidt-sebastian On M1, you may have to double check if your cmake version is 3.24.0. You'll get an error if you try to build the xcframework using any version higher than this. Linking the open issue in the opencv repo [here](https://github.com/opencv/opencv/issues/23156#issuecomment-1407375176). Not sure about Intel Mac though.

PR #4372, opened 2023-05-04
Updated iOS Image Classifier Async Callbacks to Use Delegates

priankakariatyml:
1. Updated iOS image classifier asynchronous callbacks to use delegates instead of completion blocks.
2. Added validation of packet type in MPPClassificationResult+Helpers


priankakariatyml:
Raised a PR for changing pattern of iOS image classifier async callbacks. Please see my comment [here](https://github.com/google/mediapipe/pull/4373#issuecomment-1535063344).

schmidt-sebastian (approved):

PR #4380, opened 2023-05-05
Bug Fixes in Building Opencv xcframework from Source for Cocoapods

priankakariatyml:
1. Updated podspec template  to fix MPPTasksCommon dependencies of vision and text pods to a certain version  version.
2. Updated opencv version to 4.5.3 which adds support for disabling swift
3. Updated conditional opencv xcframework dependencies in various targets


mediapipe/tasks/ios/build_ios_framework.sh
```diff
@@ -110,21 +98,24 @@ function build_ios_frameworks_and_libraries {
   # the order of a few MBs.
 
   # Build Task Library xcframework.
-  local FRAMEWORK_CQUERY_COMMAND="-c opt --apple_generate_dsym=false ${FULL_FRAMEWORK_TARGET}"
-  IOS_FRAMEWORK_PATH="$(build_target "${FRAMEWORK_CQUERY_COMMAND}")"
+  local FRAMEWORK_CQUERY_COMMAND="-c opt --apple_generate_dsym=false --define OPENCV=source ${FULL_FRAMEWORK_TARGET}"
```
priankakariatyml, line 101: The previous method of using a function to build and get the path to the output was returning the wrong path in the very first run after a bazel clean. It worked fine in subsequent runs.
Modified the logic to fix this issue.

mediapipe/tasks/ios/vision/core/BUILD
```diff
@@ -61,11 +61,5 @@ objc_library(
...
         "//third_party/apple_frameworks:UIKit",
         "@com_google_absl//absl/status:statusor",
-        "@ios_opencv//:OpencvFramework",
-    ] + select({
-        "@//third_party:opencv_ios_sim_arm64_source_build": ["@ios_opencv_source//:opencv_xcframework"],
-        "@//third_party:opencv_ios_sim_fat_source_build": ["@ios_opencv_source//:opencv_xcframework"],
-        "@//third_party:opencv_ios_arm64_source_build": ["@ios_opencv_source//:opencv_xcframework"],
-        "//conditions:default": [],
-    }),
+    ],
```
priankakariatyml, line 64: Opencv framework must be conditionally selected in framework and test targets. Linking them here caused some select config mismatches. It is also a better practice to link it to the end products since vision task runner target does not need opencv framework to build.

mediapipe/tasks/ios/MediaPipeTasksText.podspec.template
```diff
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
 
   s.module_name = 'MediaPipeTasksText'
   s.static_framework = true
-  s.dependency 'MediaPipeTasksCommon'
+  s.dependency 'MediaPipeTasksCommon', '${MPP_TASKS_COMMON_VERSION}'
```
priankakariatyml, line 15: Fixing `MPPTasksCommon` to a particular version. In case of changes in `MediaPipeTasksCommon in future versions, this will prevent any issues when installing older versions of text and vision pods.

priankakariatyml:
@schmidt-sebastian Raised a PR for fixing some issues in the Cocoapods archive building pipeline. This will make the pipeline build and link the right architectures.

schmidt-sebastian (approved):

mediapipe/tasks/ios/vision/core/BUILD
priankakariatyml, line 64: @schmidt-sebastian This line of change somehow got reversed when the PR was imported. 
This is important for the vision pod to pass CocoaPods validation. I have raised a new PR which incorporates this.
https://github.com/google/mediapipe/pull/4386

PR #4440, opened 2023-05-21
Update docs in bounding_box.d.ts

fulcus:
The BoundingBox origin point (x_origin, y_origin) is in the bottom left corner, not the top left.


schmidt-sebastian (changes requested):
I don't think this is true - https://github.com/google/mediapipe/blob/e7edd97effcc0dcdb714a45eef3590923397e6d0/mediapipe/framework/formats/location_data.proto#L50

Do you have an example that shows this problem?

PR #4464, opened 2023-05-25
Updated iOS Gesture Recogniser Result Utils

priankakariatyml:
1. Added iOS Gesture Recogniser Implementation.
2. Updated vision tasks to call new methods from vision task runner to create normalised rect.


mediapipe/tasks/ios/vision/gesture_recognizer/sources/MPPGestureRecognizer.mm
```diff
@@ -0,0 +1,273 @@
...
+                                   handednessPacket:outputPacketMap[kHandednessOutStreamName
+                                                                        .cppString]
+                                handLandmarksPacket:outputPacketMap[kLandmarksOutStreamName
+                                                                        .cppString]
+                               worldLandmarksPacket:outputPacketMap[kWorldLandmarksOutStreamName
+                                                                        .cppString]];
+}
+
+- (void)processLiveStreamResult:(absl::StatusOr<PacketMap>)liveStreamResult {
+  MPPGestureRecognizer *__weak weakSelf = self;
```
priankakariatyml, line 82: @schmidt-sebastian  We need to make a small change in the face detector.  
https://github.com/google/mediapipe/blob/master/mediapipe/tasks/ios/vision/face_detector/sources/MPPFaceDetector.mm#L223
All references to `self` inside the `processLiveStreamResult` should also be declared as `weak` not just the `self` which calls the method. Because the callback and the delegate queue will capture any reference to `self ` strongly. If not it will lead to retain cycles.
I can make this update to the face detector while getting this updated for the older vision tasks.

schmidt-sebastian: The reviewer for the FaceDetector CL suggested to use strong references as this callback is short-lived.

priankakariatyml: The callback is not short lived in memory because we pass it into the C++ task runner which stores the packets callback as an instance variable. (https://github.com/priankakariatyml/mediapipe/blob/master/mediapipe/tasks/cc/core/task_runner.h#L135).
So the packets callback and in turn the strongly referenced face detector (`self`)  will not be destroyed until the C++ task runner is destroyed. 
The face detector also has a strong reference to the C++ task runner, which will prevent the task runner from being destroyed while the detector is alive. 
This causes a retain cycle where the face detector and the C++ task runner prevent each other from being deallocated even if the user releases all strong references they have to the face detector.

Due to this, if I create` local image classifiers` in live stream mode within an infinite loop, say in like iOS's camera callbacks, I will quickly run out of memory. I tested this some time back on a device and fixed my implementation using weak references.  This was an edge case.
In reality when we create one global face detector or couple of local instances we won't run out of memory. But even in this case, the detector will never be deallocated which shouldn't happen ideally.

I will test this again to be absolutely sure. Once that is done, I will write an explanation you can take to the reviewer.

schmidt-sebastian: I will check but I think that is okay, especially since we are passing a reference to the task to the user via the delegate. So the task itself cannot be deallocated before the callback finishes running (IMHO).

schmidt-sebastian (approved):

mediapipe/tasks/ios/vision/gesture_recognizer/sources/MPPGestureRecognizer.mm
priankakariatyml: Thanks. I agree. It makes sense to retain `self`.

The only problem I see is when the user explicitly sets the global task to `nil` while the operation runs and expects it to get deallocated. With `weakSelf`, that would not happen. 

I am okay with either outcome. 

FYI, the reviewer who reviewed completion blocks did state this as one of the concerns.
So just double checking, because over here our framework itself is accessing `self` in a block which may run for some time.

schmidt-sebastian:
```
IMHO delegate calls can be considered short-lived -- the strong reference is going to exist for the life of that block, and the dispatch queue will release the block once it's finished executing.
```

We can continue to use strong references as per this.

PR #4465, opened 2023-05-25
Added utils for parsing protos from text files for iOS tests.

priankakariatyml:


schmidt-sebastian (approved):

khanhlvg:
@priankakariatyml @schmidt-sebastian  Shall we have a unit test here?

PR #4526, opened 2023-06-08
Added iOS Hand Landmarker Implementation

priankakariatyml:
1. Added iOS Hand landmarker implementation

### DIFF
--- a/mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
+++ b/mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
@@ -262,27 +262,27 @@ static NSString *const kTaskName = @"handLandmarker";
 }
 
 + (NSArray<MPPConnection *> *)handPalmConnections {
-  return kHandPalmConnections;
+  return MPPHandPalmConnections;
 }
 
 + (NSArray<MPPConnection *> *)handIndexFingerConnections {
-  return kHandIndexFingerConnections;
+  return MPPHandIndexFingerConnections;
 }
 
 + (NSArray<MPPConnection *> *)handMiddleFingerConnections {
-  return kHandMiddleFingerConnections;
+  return MPPHandMiddleFingerConnections;
 }
 
 + (NSArray<MPPConnection *> *)handRingFingerConnections {
-  return kHandRingFingerConnections;
+  return MPPHandRingFingerConnections;
 }
 
 + (NSArray<MPPConnection *> *)handPinkyConnections {
-  return kHandPinkyConnections;
+  return MPPHandPinkyConnections;
 }
 
 + (NSArray<MPPConnection *> *)handConnections {
-  return kHandConnections;
+  return MPPHandConnections;
 }
 
 @end
--- a/mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
+++ b/mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
@@ -1,0 +1,288 @@
+// Copyright 2023 The MediaPipe Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
```diff
@@ -0,0 +1,288 @@
...
+    (PacketMap &)outputPacketMap {
+  return [MPPHandLandmarkerResult
+      handLandmarkerResultWithLandmarksPacket:outputPacketMap[kLandmarksOutStreamName.cppString]
+                         worldLandmarksPacket:outputPacketMap[kWorldLandmarksOutStreamName
+                                                                  .cppString]
+                             handednessPacket:outputPacketMap[kHandednessOutStreamName.cppString]];
+}
+
+- (void)processLiveStreamResult:(absl::StatusOr<PacketMap>)liveStreamResult {
+  if (![self.handLandmarkerLiveStreamDelegate
```
schmidt-sebastian, line 76: FYI I got feedback in an earlier CL to make the `didFinishDetectionWithResult` method in the protocol non-optional and instead make the delegate optional. This is meant to make implementing the delegate less error-prone. Can you update the PR?

mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarksConnections.h
```diff
@@ -0,0 +1,58 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import "mediapipe/tasks/ios/components/containers/sources/MPPConnection.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSArray<MPPConnection *> *const kHandPalmConnections = @[
```
schmidt-sebastian, line 20: "Prefixes are commonly required in Objective-C to avoid naming collisions in a global namespace. Classes, protocols, global functions, and global constants should generally be named with a prefix that begins with a capital letter followed by one or more capital letters or numbers."

vs

"A lowercase k can be used as a standalone prefix for constants of static storage duration declared within implementation files:
NOTE: Previous convention was for public constant names to begin with a lowercase k followed by a project-specific [prefix](https://g3doc.corp.google.com/company/teams/objc-readability/style_guide.md?cl=head#prefixes). This practice is no longer recommended."

These are pretty close to being global constants, so we should probably follow this guideline.

mediapipe/tasks/ios/vision/hand_landmarker/sources/MPPHandLandmarker.mm
priankakariatyml: The delegate is already optional. But we don't let the task get initialised if user doesn't set it in the `MPPVisionTaskRunner`. I'll raise separate PRs for this since it will affect all tasks and the tests too.

The method is optional because in future we will also add a method with an extra argument for returning the source MPPImage. So users can implement either one of them based on what they prefer. Even if I change it now, it may end up getting updated later.
Not uncommon to have protocols with only optionals methods. The following protocol delivers camera buffers in iOS.
https://developer.apple.com/documentation/avfoundation/avcapturevideodataoutputsamplebufferdelegate?language=objc

LMK what you think.

priankakariatyml: Also when you remove the mandatory delegate check we have either one of these options:
1. Provide a C++ packets callback irrespective. Otherwise users won't be able to call `detectAsyncInImage()` without facing callback errors from C++.
2. Not provide a packets callback and store a flag in vision task runner to indicate the same. Use this flag to return a better error message to users when `processLiveStreamPacketMap()` is called since the C++ error would not be meaningful to the iOS clients.
I prefer option 2 as this would avoid any unnecessary processing.

schmidt-sebastian: Thanks for the explanation. Do you think it makes sense to have two different delegate protocols, each with their own non-optional method? This might be more cumbersome than having a single delegate with optional methods but I just wanted to raise the possibility. The reviewer was concerned that there is very little compile time safety for optional protocols (e.g. if a user mistypes a method name, nothing happens). I will defer to your judgment here. Please do update the FaceLandmarker and FaceDetector APIs as well so we have consistency.

As for your last question, I have a strong preference for option 2.

schmidt-sebastian (approved):

PR #4538, opened 2023-06-13
Added iOS segmentation mask

priankakariatyml:


mediapipe/tasks/ios/vision/core/sources/MPPMask.h
```diff
@@ -0,0 +1,111 @@
...
+ * width and height.
+ *
+ * @param uint8Data A pointer to the memory location of the `UInt8` data array.
+ * @param width The width of the mask.
+ * @param height The height of the mask.
+ *
+ * @return A new `MPPMask` instance with the given `UInt8*` data, width and height.
+ */
+- (nullable instancetype)initWithUInt8Data:(const UInt8 *)uint8Data
+                                     width:(NSInteger)width
```
khanhlvg, line 82: Since you store the width and height as CGFloat, why don't having the initializer to accept CGFloat for consistency?

priankakariatyml: My bad. It should be NSInteger. Updated this.

mediapipe/tasks/ios/vision/core/sources/MPPMask.mm
```diff
@@ -0,0 +1,157 @@
...
+                                       width:(NSInteger)width
+                                      height:(NSInteger)height {
+  self = [self initWithWidth:width height:height dataType:MPPMaskDataTypeFloat32 error:nil];
+  if (self) {
+    _float32Data = float32Data;
+  }
+  return self;
+}
+
+- (instancetype)initWithUInt8DataToCopy:(const UInt8 *)uint8DataToCopy
```
schmidt-sebastian, line 82: Can we combine this and `initWithUInt8Data` and pass a flag to determine whether ownership should be taken?

mediapipe/tasks/ios/vision/core/sources/MPPMask.mm
```diff
@@ -0,0 +1,157 @@
...
+                                 height:(NSInteger)height {
+  self = [self initWithWidth:width height:height dataType:MPPMaskDataTypeUInt8 error:nil];
+  if (self) {
+    _uint8Data = allocateDataPtr(_allocatedUInt8Data, _width * _height);
+    copyData(_uint8Data, uint8DataToCopy, _width * _height);
+  }
+  return self;
+}
+
+- (instancetype)initWithFloat32DataToCopy:(const float *)float32DataToCopy
```
schmidt-sebastian, line 93: Same comment here.

mediapipe/tasks/ios/vision/core/sources/MPPMask.h
```diff
@@ -0,0 +1,110 @@
...
+@property(nonatomic, readonly) MPPMaskDataType dataType;
+
+/**
+ * The pointer to the memory location where the underlying mask as a single channel `UInt8` array is
+ * stored.
+ */
+@property(nonatomic, readonly, assign) const UInt8 *uint8Data;
+
+/**
+ * The pointer to the memory location where the underlying mask as a single channel float 32 array
```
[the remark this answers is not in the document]
priankakariatyml: Updated this to `32 bit float` because there is no `Float32` type as such.

mediapipe/tasks/ios/vision/core/sources/MPPMask.mm
```diff
@@ -0,0 +1,157 @@
...
+template <typename T>
+void copyData(const T *destination, const T *source, size_t length) {
+  memcpy((void *)destination, source, length * sizeof(T));
+}
+}  // namespace
+
+@interface MPPMask () {
+  const UInt8 *_uint8Data;
+  const float *_float32Data;
+  std::unique_ptr<UInt8[]> _allocatedUInt8Data;
```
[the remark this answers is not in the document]
priankakariatyml: Made it `uInt8DataPtr`

mediapipe/tasks/ios/vision/core/sources/MPPMask.mm
```diff
@@ -0,0 +1,157 @@
...
+// limitations under the License.
+
+#import "mediapipe/tasks/ios/vision/core/sources/MPPMask.h"
+#import "mediapipe/tasks/ios/common/sources/MPPCommon.h"
+#import "mediapipe/tasks/ios/common/utils/sources/MPPCommonUtils.h"
+
+namespace {
+template <typename T>
+T *allocateDataPtr(std::unique_ptr<T[]> &data, size_t length) {
+  data = std::unique_ptr<T[]>(new T[length]);
```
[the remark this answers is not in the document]
priankakariatyml: Only uninitialised `unique_ptr`  instance variables are passed by reference to this call from within this file. So no values are discarded. They are initialised from within the function and the changes are visible outside the function scope too due to the pass by reference. Also unique_ptr.get() returns a pointer, which would be valid for the scope of the ptr.
Just saves a couple of lines of code  and also the need to specify the type for alloc and copy. 

Technically, if the passed in ptr is already initialised, it would discard the existing value.  To make it fool proof for initialised ptrs, we need to do a null check which is unnecessary here. Hence dropped it.

Nevertheless, I have added these lines in place now to avoid any further pushback during the ObjC review.

mediapipe/tasks/ios/vision/core/sources/MPPMask.mm
priankakariatyml: We can. But should we expose this option in the interface ?

schmidt-sebastian: I don't think it hurts. The behavior is easy to explain and it might make sense for external use cases as well (e.g. tests).

priankakariatyml: Fixed

mediapipe/tasks/ios/vision/core/sources/MPPMask.h
```diff
@@ -0,0 +1,118 @@
...
+  /** Represents the native `float *` type. */
+  MPPMaskDataTypeFloat32,
+
+} NS_SWIFT_NAME(MaskDataType);
+
+/**
+ * The wrapper class for MediaPipe segmentation masks.
+ *
+ * Masks are stored as `UInt8 *` or `float *` objects.
+ * Every mask is has an underlying type which can be accessed using `dataType`. You can access the
```
schmidt-sebastian, line 34: Typoe

mediapipe/tasks/ios/vision/core/sources/MPPMask.h
```diff
@@ -0,0 +1,118 @@
...
+ * mask as any other type using the appropriate properties. For eg:, if the underlying type is
+ * `MPPMaskDataTypeUInt8`, in addition to accessing the mask using `uint8Array`, you can access
```
schmidt-sebastian, line 36: Missing "it" or "the data".

We should mention the different value ranges as well.

schmidt-sebastian (approved):

PR #4583, opened 2023-06-30
Removed testing for delegates in different running modes from iOS Vision Tasks

priankakariatyml:


mediapipe/tasks/ios/core/sources/MPPTaskRunner.mm
```diff
@@ -59,7 +62,20 @@ - (instancetype)initWithCalculatorGraphConfig:(CalculatorGraphConfig)graphConfig
 }
 
 - (BOOL)sendPacketMap:(const PacketMap &)packetMap error:(NSError **)error {
+  if (!_initializedWithPacketsCallback) {
+    [MPPCommonUtils
+        createCustomError:error
+                 withCode:MPPTasksErrorCodeFailedPreconditionError
+              description:[NSString
+                              stringWithFormat:@"This method can only be called if the task is "
+                                               @"running in a stream mode and an object of a class "
```
schmidt-sebastian, line 71: "an object of a class" is a bit hard to parse. Maybe just "and a delegate is provided in the task options..."

mediapipe/tasks/ios/core/sources/MPPTaskRunner.mm
```diff
@@ -59,7 +62,20 @@ - (instancetype)initWithCalculatorGraphConfig:(CalculatorGraphConfig)graphConfig
 }
 
 - (BOOL)sendPacketMap:(const PacketMap &)packetMap error:(NSError **)error {
+  if (!_initializedWithPacketsCallback) {
```
schmidt-sebastian, line 65: This now means that we can initialize a task that will only ever throw errors. Is it possible to revert to the old behavior and throw an error on the initialization?

priankakariatyml: Yeah, but then this whole change is unnecessary right? I mean wouldn't it suffice to just make all the delegate methods @required. Can I close this PR ?

schmidt-sebastian: Probably. I think we can have `@required` methods and tests that the delegate object is set itself (like I believe we do today).

PR #4590, opened 2023-07-03
Added iOS Image Segmenter Result and Options Helpers

priankakariatyml:


mediapipe/tasks/ios/vision/image_segmenter/utils/sources/MPPImageSegmenterResult+Helpers.h
```diff
@@ -0,0 +1,48 @@
...
+ * Creates an `MPPImageSegmenterResult` from confidence masks, category mask and quality scores
+ * packets.
+ *
+ * If `shouldCopyMaskPacketData` is set to `YES`, the confidence and catergory masks of the newly
+ * created `MPPImageSegmenterResult` holds references to deep copied pixel data of the output
+ * respective masks.
+ *
+ * @param confidenceMasksPacket A MediaPipe packet wrapping a `std::vector<mediapipe::Image>`.
+ * @param categoryMaskPacket A MediaPipe packet wrapping a `<mediapipe::Image>`.
+ * @param qualityScoresPacket a MediaPipe packet wrapping a `std::vector<float>`.
```
schmidt-sebastian, line 33: s/a/A/ for consistency

mediapipe/tasks/ios/vision/image_segmenter/utils/sources/MPPImageSegmenterResult+Helpers.h
```diff
@@ -0,0 +1,48 @@
...
+ * @param shouldCopyMaskPacketData A `BOOL` which indicates if the pixel data of the output masks
```
schmidt-sebastian, line 34: Do you think this will be needed?

priankakariatyml: Yeah we have two sets of APIs for each running mode as per Java. Image and video have apis which return a reference or copy of the underlying pixel buffer. So based on which API is invoked, we need to pass this info to MPMask when it is initialised to take appropriate action.

schmidt-sebastian (approved):

PR #4645, opened 2023-07-26
Added refactored iOS vision task runner sources

priankakariatyml:


schmidt-sebastian (changes requested):
Do you mind fixing this?

```
convenience initializer should not invoke an initializer on 'super' [-Werror,-Wobjc-designated-initializers]
  100 |   self = [super initWithCalculatorGraphConfig:graphConfig
      |                 ^

```

This turns out to be more involved that I thought (probably not super complicated, but it might be better to update the PR).

priankakariatyml:
> convenience initializer should not invoke an initializer on 'super'



> Do you mind fixing this?
> 
> ```
> convenience initializer should not invoke an initializer on 'super' [-Werror,-Wobjc-designated-initializers]
>   100 |   self = [super initWithCalculatorGraphConfig:graphConfig
>       |                 ^
> ```
>
> This turns out to be more involved that I thought (probably not super complicated, but it might be better to update the PR).

I thought this PR was approved. Sorry, I didn't notice this. I have fixed it. Removed `initWithCalculatorGraphConfig`. We need only one designated initialiser which does everything

schmidt-sebastian (approved):

PR #4702, opened 2023-08-17
Add undefined check to 'document' variable in platform_utils

spennihana:
It's possible that `document` is not defined, e.g., when we're operating in a web worker. This PR adds an `undefined` check here and makes `isIOS` return false whenever it's not defined.


spennihana:
Closing this out, thanks @schmidt-sebastian for already having a fix ready to go: 990bfd2e3eaf0f0c977a7af7330ac4ad54f768cb

PR #4766, opened 2023-09-06
Added iOS Image Segmenter ObjC Tests

priankakariatyml:
1. Added iOS Image Segmenter ObjC Tests.
2. Moved utilities for creating test resource files to helpers.
3. Fixed a bug due to a typo in MPPImageSegmenterResult helper.


mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
```diff
@@ -0,0 +1,234 @@
...
+    multipliedMask[i] = mask1[i] * mask2[i];
+  }
+
+  return multipliedMask;
+}
+
+double softIOU(const float *mask1, const float *mask2, size_t size) {
+  float *interSectionVector = multiply(mask1, mask2, size);
+  double interSectionSum = sum(interSectionVector, size);
+  free(interSectionVector);
```
schmidt-sebastian, line 61: The memory ownership is hard to follow here. Please allocate the memory here, and pass the pointer to "multiply". You can also use a `vector` and avoid manual memory management altogether.

mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
```diff
@@ -0,0 +1,234 @@
...
+- (void)assertResultsOfSegmentImageWithFileInfo:(MPPFileInfo *)imageFileInfo
+                                         usingImageSegmenter:(MPPImageSegmenter *)imageSegmenter
+    approximatelyEqualsExpectedCategoryMaskImageWithFileInfo:
+        (MPPFileInfo *)expectedCategoryMaskFileInfo
+                                   shouldHaveConfidenceMasks:(BOOL)shouldHaveConfidenceMasks {
+  MPPImageSegmenterResult *result = [self segmentImageWithFileInfo:imageFileInfo
+                                               usingImageSegmenter:imageSegmenter];
+
+  XCTAssertNotNil(result.categoryMask);
+  shouldHaveConfidenceMasks ? XCTAssertNotNil(result.confidenceMasks)
```
schmidt-sebastian, line 156: Please use a proper if-statement.

schmidt-sebastian (changes requested):

mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
priankakariatyml: The operations are performed on float masks from copied `MPPMask`s. I have an added an `MPPMask` util category to instantiate a copied MPPMask from a golden mask image file for the tests. https://github.com/google/mediapipe/pull/4766/files#diff-305b13f38edec0661995cf9ebab815d2c2600ac8b2df20520571b335e57825a9
Its [float32Data](https://github.com/priankakariatyml/mediapipe/blob/master/mediapipe/tasks/ios/vision/core/sources/MPPMask.h#L68) is compared with the float buffer of the copied `MPPMask` that the image  segmenter returns. This won't be a problem in async tests too since we will perform this calculation in the delegate method or the closure which returns the result. 

IMO, one more copy here is redundant. Would moving these to the method that performs confidence mask comparison improve readability like for category masks? Code was longer, hence moved it to separate methods. 
Or we can also change the arguments to `MPPMask`. So it becomes evident that memory management is handled in the outer function or within the mask.

schmidt-sebastian: I don't mind the extra copy since this is just test code. What I am a bit concerned about is that the consumer of "multiply()" needs to know that multiply allocates new memory on the heap so that it can be freed. This is not apparent without looking at the code and might lead to memory leaks in the future (again, only in tests, but we should still try to avoid this). 

I think the simplest way to fix this is to return a vector:

````
std::vector<float> multiply(const float *mask1, const float *mask2, size_t size) {
  double sum = 0.0;
  std::vector<float> multipliedMask{size};
  for (int i = 0; i < size; i++) {
    multipliedMask[i] = mask1[i] * mask2[i];
  }
  return multipliedMask;
}
```

Otherwise, this function should ideally be:

```
std::vector<float> multiply(const float *mask1, const float *mask2, float* output_mask, size_t size) {...}

{
...
float *multipliedMask = (float *)malloc(size * sizeof(float));
multiply(m1, m2, multipliedMask, size);
free(multipliedMask);
...
}
```

Both make it very clear at the callsite what needs to be done. thanks.

priankakariatyml: My bad. I mistook your comment to be about the allocation of the args `mask1` and `mask2` of `softIOU()` instead of the multiplied mask.  Hence my concern about the extra copy. 

I have updated this to an std::vector in my local. I can't push to this PR at the moment due to some issue with our commit email ids. Hence the CLA check is failing.

Once the admin team resolves the problem, I'll push the fixes.

priankakariatyml:
Fixed all issues in new PR.
https://github.com/google/mediapipe/pull/4802/files

PR #4771, opened 2023-09-08
Adds delegate to enable GPU acceleration in MPPVisionTask

col-fi:
Added GPU acceleration delegate setting.
Directly created `Packet<Image>` without converting to `ImageFrame`, as `mediapipe::Image` stores `CVPixelBufferRef` as  `GPUBuffer`.


schmidt-sebastian:
Thank you for sending this. This has just been addressed internally: https://github.com/google/mediapipe/pull/4924

PR #4793, opened 2023-09-13
Added Face Stylizer unit test files

kinaryml:


mediapipe/tasks/python/vision/face_stylizer.py
```diff
@@ -54,11 +55,15 @@ class FaceStylizerOptions:
   """
 
   base_options: _BaseOptions
+  running_mode: _RunningMode = _RunningMode.IMAGE
```
khanhlvg, line 58: We don't support RunningMode in FaceStylizer. It must always be .IMAGE

kinaryml: Fixed it now.

PR #4818, opened 2023-09-20
Adds an empty skeleton project for iOS docgen.

markmcd:
Dependencies tracked in `Podfile` are used to generate reference docs.

LMK if there's a better place for this.


docs/MediaPipeTasksDocGen/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '15.0'
+
+target 'MediaPipeTasksDocGen' do
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
+
+  # Pods for MediaPipeTasksDocGen
+  pod 'MediaPipeTasksText', '0.10.5'
```
schmidt-sebastian, line 9: Would it be better not to pin the versions here? What's the process for updating the docs?

schmidt-sebastian (approved):

docs/MediaPipeTasksDocGen/Podfile
markmcd: Honestly I don't know - do you ever do beta or rc releases? If so, would they come through in a pod dep?

For other platforms & tools we always want to document the latest, stable release, and scripts are often used in automation (though iOS does not do this yet) where they run nightly.

If unpinning the version number achieves latest-stable without any nightly or RC releases getting through then that sounds good.

markmcd: OK I did some research. Looks like Cocoapods follows semver and won't install RC releases ahead of stable.

Versions are unpinned now so I'll merge this before I go OOO.

Thanks @schmidt-sebastian !

markmcd: This OK to merge now?

schmidt-sebastian (approved):

PR #4830, opened 2023-09-22
Added tests for iOS Image Segmenter methods with completion handlers

priankakariatyml:


mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
```diff
@@ -468,6 +468,45 @@ - (void)imageSegmenter:(MPPImageSegmenter *)imageSegmenter
   }
```
schmidt-sebastian, line 468: Can this method be moved to "Assert Segmenter Results"?

mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
```diff
@@ -468,6 +468,45 @@ - (void)imageSegmenter:(MPPImageSegmenter *)imageSegmenter
...
+#pragma mark Mask No Copy Tests
+
+- (void)testSegmentWithNoCopyConfidenceMasksAndImageModeSucceeds {
+  MPPImageSegmenterOptions *options =
+      [self imageSegmenterOptionsWithModelFileInfo:kImageSegmenterModelFileInfo];
+
+  MPPImageSegmenter *imageSegmenter = [self createImageSegmenterWithOptionsSucceeds:options];
+  
+  MPPImage *image = [MPPImage imageWithFileInfo:kCatImageFileInfo];
+  [imageSegmenter segmentImage:image withCompletionHandler:^(MPPImageSegmenterResult *result, NSError *error) {
```
schmidt-sebastian, line 480: Don't these test need expectation to ensure that these async blocks execute?

priankakariatyml: No. Because the method returns only after the completion handler call completes and hence the test gets completed only after the result is returned via the handler. This segment method is synchronous as opposed to the live stream segment method.

mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
schmidt-sebastian: I think this method should be moved out of the section that contains the test methods. I will move it on import.

mediapipe/tasks/ios/test/vision/image_segmenter/MPPImageSegmenterTests.mm
schmidt-sebastian: Okay, that's right. I will add a comment so future me does not forget again.

priankakariatyml:
@schmidt-sebastian Just a reminder that this PR is not yet merged.

schmidt-sebastian:
> @schmidt-sebastian Just a reminder that this PR is not yet merged.

Oops :)

PR #4840, opened 2023-09-28
Added method to create iOS `MPImage` from C++ `Image`.

priankakariatyml:
1. Added a method to create iOS `MPImage` from C++ `Image`.
2. Fixed incorrect R, G, B values being passed in for non-opaque images.

### DIFF
--- a/mediapipe/tasks/ios/vision/core/utils/sources/MPPImage+Utils.mm
+++ b/mediapipe/tasks/ios/vision/core/utils/sources/MPPImage+Utils.mm
@@ -25,17 +25,27 @@
 #include "mediapipe/framework/formats/image_format.pb.h"
 
 namespace {
+using ::mediapipe::Image;
+using ::mediapipe::ImageFormat;
 using ::mediapipe::ImageFrame;
+
+vImage_Buffer allocatedVImageBuffer(vImagePixelCount width, vImagePixelCount height,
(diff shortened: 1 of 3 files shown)


mediapipe/tasks/ios/vision/core/utils/sources/MPPImage+Utils.mm
```diff
@@ -214,29 +210,91 @@ @implementation MPPCGImageUtils
...
+
+  switch (internalImageFrame->Format()) {
+    case ImageFormat::SRGBA: {
+      bitmapInfo = kCGImageAlphaPremultipliedLast | kCGBitmapByteOrder32Big;
+      break;
+    }
+    default:
+      [MPPCommonUtils createCustomError:error
+                               withCode:MPPTasksErrorCodeInternalError
+                            description:@"Unsupported Image Format Conversion."];
```
schmidt-sebastian, line 245: Can you add the provided format name?

priankakariatyml: I have changed this to: "Some internal error occurred." I think this is more apt since the function converts an output image frame to an iOS image for tasks like the face styliser. Most likely this won't happen, unless there is some issue in the library source code.

schmidt-sebastian (approved):

PR #4948, opened 2023-11-09
fix: start graph for the case where the input sourceMode is set to video

sabade-omkar:
If the `sourceMode` is set to `MediaPipeDemoSourceVideo`, a crash occurs due to the graph not being initiated. This commit includes a fix for the issue.


mediapipe/examples/ios/common/CommonViewController.mm
```diff
@@ -113,7 +113,7 @@ - (void)viewWillAppear:(BOOL)animated {
       self.videoSource = [[MPPPlayerInputSource alloc] initWithAVAsset:video];
       [self.videoSource setDelegate:self queue:self.videoQueue];
       dispatch_async(self.videoQueue, ^{
-        [self.videoSource start];
+        [self startGraphAndVideo];
```
schmidt-sebastian, line 116: Can you align this with "startGraphAndCamera" and move the dispatch block to the new function? Thanks

schmidt-sebastian (approved):

PR #4954, opened 2023-11-11
MediaPipe Tasks Python Image Classifier Benchmark

kinaryml:


mediapipe/tasks/python/benchmark/vision/image_classifier/image_classifier_benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+import numpy as np
+import mediapipe as mp
+from mediapipe.tasks import python
+from mediapipe.tasks.python import vision
+
+_IMAGE_FILE = 'burger.jpg'
+
+
+def run(model: str, n_iterations: int, delegate: python.BaseOptions.Delegate):
+    """Run asynchronous inference on images and benchmark.
```
schmidt-sebastian, line 26: Please document the return value (and maybe allow us to pass in the percentile to use?)

kinaryml: I've fixed it now.

schmidt-sebastian (approved):

mediapipe/tasks/python/benchmark/vision/image_classifier/setup.sh
```diff
@@ -0,0 +1,6 @@
+# Install Python dependencies.
```
schmidt-sebastian, line 1: Note that I removed this file as part of the import. I changed your script to build via Bazel (added a BUILD file). I moved the commands to download the models to the README. Building via Bazel allows us to ensure that these script don't break as we change our CI.

kinaryml: Agreed and it ensures we're pulling the updated models while benchmarking.

PR #4963, opened 2023-11-15
Fixed GPU Delegate Crash on iOS Simulator

priankakariatyml:
1. Fixed crash in creation of OpenGL ES texture from CVPixelBuffer on iOS simulator
2. Fixed unsupported Metal texture creation from a Metal Buffer on iOS Simulator


schmidt-sebastian (approved):

schmidt-sebastian:
Note that this is causing some issues with internal tests, which I need to fix before merging.

PR #4993, opened 2023-11-29
Gesture Recognizer C Tasks API

kinaryml:


kinaryml:
@schmidt-sebastian I've resolved the comments. PTAL and lmk.

mediapipe/tasks/c/components/containers/gesture_recognizer_result_converter.cc
```diff
@@ -47,18 +47,15 @@ void CppConvertToGestureRecognizerResult(
...
-        cpp_category.category_name = classification.label();
-      }
-      if (classification.has_display_name()) {
-        cpp_category.display_name = classification.display_name();
-      }
+      cpp_category.index =
+          classification.has_index() ? classification.index() : 0;
+      cpp_category.score =
+          classification.has_score() ? classification.score() : 0.0f;
+      cpp_category.category_name =
```
schmidt-sebastian, line 54: These should be initialized to an empty optional.

mediapipe/tasks/c/components/containers/gesture_recognizer_result_converter.cc
```diff
@@ -47,18 +47,15 @@ void CppConvertToGestureRecognizerResult(
...
-        cpp_category.score = classification.score();
-      }
-      if (classification.has_label()) {
-        cpp_category.category_name = classification.label();
-      }
-      if (classification.has_display_name()) {
-        cpp_category.display_name = classification.display_name();
-      }
+      cpp_category.index =
+          classification.has_index() ? classification.index() : 0;
```
schmidt-sebastian, line 51: The has_index() check is not needed since 0 is the default value.

schmidt-sebastian (approved):
Thanks! I'll resolve my new comments during import.

mediapipe/tasks/c/components/containers/BUILD
```diff
@@ -209,3 +236,28 @@ cc_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+cc_library(
```
schmidt-sebastian, line 240: I noticed this during import, but since this is not a shared utility this should live in the task's folder. I am moving this.

I will also move the lanugage_detector_result_converter.

schmidt-sebastian (approved):

PR #5007, opened 2023-12-04
Added iOS Pose Landmarker to CocoaPods build

priankakariatyml:


priankakariatyml:
@schmidt-sebastian Could you take a look at this PR ?

schmidt-sebastian (approved):

WaterBoiledPizza:
Is Pose Landmarker fully available for iOS? Will there be an example of how to set it up eventually? Thank you.

schmidt-sebastian:
Yes - it is available. We don't have an example yet but will add one soon.

PR #5021, opened 2023-12-11
Face Landmarker C Tasks API and tests

kinaryml:


mediapipe/tasks/c/components/containers/matrix_converter.cc
```diff
@@ -0,0 +1,45 @@
...
+
+namespace mediapipe::tasks::c::components::containers {
+
+void CppConvertToMatrix(const Eigen::MatrixXf& in, ::Matrix* out) {
+  out->rows = in.rows();
+  out->cols = in.cols();
+  out->data = new float[out->rows * out->cols];
+
+  // Copy data from Eigen matrix to C matrix in column-major order
+  for (int col = 0; col < out->cols; ++col) {
```
schmidt-sebastian, line 31: Can we avoid the one by one copy here by using https://eigen.tuxfamily.org/dox/classEigen_1_1PlainObjectBase.html#ac54123f62de4c46a9107ff53890b6116 ? Note that the order of elements does not really matter here as long as we document it.

mediapipe/tasks/c/vision/face_landmarker/face_landmarker_result.h
```diff
@@ -0,0 +1,59 @@
...
+#endif  // MP_EXPORT
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// The hand landmarker result from HandLandmarker, where each vector
+// element represents a single hand detected in the image.
+struct FaceLandmarkerResult {
+  // Optional face blendshapes results.
```
schmidt-sebastian, line 36: Nit: Since this is optional, can you move this down in the struct?

mediapipe/tasks/c/vision/face_landmarker/face_landmarker_result.h
```diff
@@ -0,0 +1,59 @@
...
+  // The number of elements in the face_blendshapes array.
+  uint32_t face_blendshapes_count;
+
+  // Detected face landmarks in normalized image coordinates.
+  struct NormalizedLandmarks* face_landmarks;
+
+  // The number of elements in the face_landmarks array.
+  uint32_t face_landmarks_count;
+
+  // Optional facial transformation matrix.
```
schmidt-sebastian, line 48: matrixes

mediapipe/tasks/c/vision/face_landmarker/face_landmarker_test.cc
```diff
@@ -0,0 +1,291 @@
...
+constexpr float kLandmarksPrecision = 0.03;
+constexpr float kBlendshapesPrecision = 0.12;
+constexpr float kFacialTransformationMatrixPrecision = 0.05;
+constexpr int kIterations = 100;
+
+std::string GetFullPath(absl::string_view file_name) {
+  return JoinPath("./", kTestDataDirectory, file_name);
+}
+
+void MatchesFaceLandmarkerResult(FaceLandmarkerResult* result,
```
schmidt-sebastian, line 51: s/Matches/Assert

schmidt-sebastian (changes requested):

mediapipe/tasks/c/components/containers/matrix_converter.cc
kinaryml: @schmidt-sebastian I've documented it now and simplified the copying process.

mediapipe/tasks/c/components/containers/matrix_converter.cc
```diff
@@ -0,0 +1,47 @@
...
+  out->rows = in.rows();
+  out->cols = in.cols();
+  out->data = new float[out->rows * out->cols];
+
+  // Copy data from Eigen matrix to C-style matrix.
+  // This operation copies the elements sequentially as they appear in the Eigen
+  // matrix's internal storage, regardless of whether it's stored in row-major
+  // or column-major order and ensures the integrity of data during the
+  // transfer.
+  for (int i = 0; i < out->rows * out->cols; ++i) {
```
schmidt-sebastian, line 35: Two comments:
- You should be able to use memcpy now.
- You also need to add a comment in the Matrix type to indicate how data is stored. Please verify, but I believe that Eigen uses column-major (you can verify via unit test).

kinaryml: @schmidt-sebastian I've used memcpy now and I think I've already mentioned the Matrix stores the data in column-major order already.

schmidt-sebastian (approved):

PR #5052, opened 2024-01-05
Pose Landmarker C Tasks API

kinaryml:


mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker.h
```diff
@@ -0,0 +1,149 @@
...
+  // The user-defined result callback for processing live stream data.
+  // The result callback should only be specified when the running mode is set
+  // to RunningMode::LIVE_STREAM. Arguments of the callback function include:
+  // the pointer to recognition result, the image that result was obtained
+  // on, the timestamp relevant to recognition results and pointer to error
+  // message in case of any failure. The validity of the passed arguments is
+  // true for the lifetime of the callback function.
+  //
+  // A caller is responsible for closing pose landmarker result.
+  typedef void (*result_callback_fn)(PoseLandmarkerResult* result,
```
schmidt-sebastian, line 74: This should be a const reference (result).

We also still need to update the other tasks to use the same type.

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_result_converter.cc
```diff
@@ -0,0 +1,97 @@
...
+  for (uint32_t i = 0; i < result->pose_world_landmarks_count; ++i) {
+    CppCloseLandmarks(&result->pose_world_landmarks[i]);
+  }
+  delete[] result->pose_world_landmarks;
+
+  result->pose_landmarks = nullptr;
+  result->pose_world_landmarks = nullptr;
+
+  result->pose_landmarks_count = 0;
+  result->pose_world_landmarks_count = 0;
```
schmidt-sebastian, line 94: Need to free the masks here as well.

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_test.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  const auto& image_frame = image->GetImageFrameSharedPtr();
+  const MpImage mp_image = {
+      .type = MpImage::IMAGE_FRAME,
+      .image_frame = {.format = static_cast<ImageFormat>(image_frame->Format()),
+                      .image_buffer = image_frame->PixelData(),
+                      .width = image_frame->Width(),
+                      .height = image_frame->Height()}};
+
+  for (int i = 0; i < kIterations; ++i) {
```
schmidt-sebastian, line 138: I don't think we gain much by running this more than 2 times. What do you think?

(The live stream test is different since it tests flow limits).

schmidt-sebastian (changes requested):

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_test.cc
kinaryml: The C++ test does it so thought I'd do it here too.

https://github.com/google/mediapipe/blob/495012645626e1b1e4da98a67919c53ac7f1520f/mediapipe/tasks/cc/vision/pose_landmarker/pose_landmarker_test.cc#L386

We can cut it down since we already know it runs well in C++ perhaps?

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_result_converter.cc
kinaryml: Since segmentation_masks[i].image_frame.image_buffer is not dynamically allocated here, we do not delete it in the cleanup, right?

schmidt-sebastian: I think you still need to clear https://github.com/google/mediapipe/pull/5052/files/cc0d3466afaba965520ae9ac2578eefb2da3e5c9#diff-497245f04089ed54cca2729d05cf2d0be6c370d1d2072d5c4b0a606345cb22bdR38 ?

kinaryml: @schmidt-sebastian I've added it now, but I get a SIGSEGV when I try to use free or delete on `const uint8_t* image_buffer`. Could you take a look? I believe casting away its constness and deleting it can lead to undefined behavior

schmidt-sebastian: I'll debug during import.

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_test.cc
schmidt-sebastian: Yeah - that's what I am thinking. Will set to 10 during import.

schmidt-sebastian (approved):

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_test.cc
schmidt-sebastian: ( I see you already set it to 5 - that's good as well. So keeping this).

mediapipe/tasks/c/vision/pose_landmarker/pose_landmarker_result_converter.cc
schmidt-sebastian: You actually only need to free the image array and not the data itself since the data is statically allocated (which could present a problem depending on how the API is used, but this is outside of the scope of this PR).

Issue #4605, opened 2023-07-10
Pose Landmarker returns NormalizedLandmarks objects without visibility


mihaiEDW:
### MediaPipe Solution (you are using)

Pose Landmark Detection

### Programming language

typescript

### Are you willing to contribute it

Yes

### Describe the feature and the current behaviour/state

```shell
Current behaviour:

The documentation states here (https://developers.google.com/mediapipe/solutions/vision/pose_landmarker/web_js#handle_and_display_results) that the Pose Landmarker returns a poseLandmarkerResult object, which in turn contains a visibility parameter.

However, when using the demo provided by the Mediapipe Team, there is no such visibility parameter returned.

Feature:

I'd like to propose a small change to the information returned in the NormalizedLandmarks object: visibility and presence, as per the current documentation and the past Mediapipe versions.
```

### Will this change the current API? How?

```shell
Yes, the change will be visible in the NormalizedLandmarks object, as it will contain two more parameters when returned: visibility and presence.

This will not only allow developers use the current version of Mediapipe in the same ways that they used it in the past, but it will allow for extended functionality, for a very small change.
```

### Who will benefit with this feature?

The developers building apps and solutions using Mediapipe.

### Please specify the use cases for this feature

```shell
- view if the people are fully in the camera frame
- view which pose landmarks are seen correctly
- view how the person is standing, relative to the camera
```

### Any Other info

```shell
It seems that this feature was available in the past versions and I can't understand why it is no more available.
```

kuaashish:
Hello @lu-wang-g,

Could you please look into this feature request. Thank you

galharth:
Any update ? thanks :)

ewan-m:
I am interested in this also! Happy to contribute a PR for it (visibility and presence both) 😄

schmidt-sebastian:
@yichunk - We need to take another look.

kuaashish:
Hi @mihaiEDW,

Could you please verify the current status using the nightly version at https://www.npmjs.com/package/@mediapipe/tasks-vision/v/0.10.12-rc.20240407? It must return the visibility due to the merged [pull request](https://github.com/google/mediapipe/pull/5142). These changes are expected to be included in the upcoming major release.

Thank you!!

sarelpoy:
for now, I found a way around it:
I took tasks-vision.js(https://cdn.skypack.dev/@mediapipe/tasks-vision@0.10.0) from my server and edited it to add visibility to the results
```
[...]

Issue #4409, opened 2023-05-15
Mediapipe Web Pose Landmarker doesn't return visibility 


jays0606:
### Have I written custom code (as opposed to using a stock example script provided in MediaPipe)

None

### OS Platform and Distribution

Mac OS

### Task name (e.g. Image classification, Gesture recognition etc.)

Pose Estimation

### Programming Language and version (e.g. C++, Python, Java)

JavaScript

### Describe the actual behavior

```shell
It says in the doc mediapipe pose landmark should return visibility, but in reality the output only returns x, y, z.
```

### Describe the expected behaviour

```shell
The visibility 0.0 ~ 1.0 should be returned along with xyz.
```

### Standalone code/steps you may have used to try to get what you need

```shell
const poseLandmarker = await PoseLandmarker.createFromOptions(vision, {
        baseOptions: {
            modelAssetPath: `pose_landmarker_full.task`,
            delegate: "GPU",
        },
        runningMode: "VIDEO",
        numPoses: 1,
    });

And run poseLandmark.detectForVideos()
I printed out the results.
```

Bilalismail:
Facing the same issue. I'm currently using Z values to determine visibility by scoring against right and left landmarks. Can you confirm if the visibility score will be coming back?

emoc11:
Same issue here, could be greate to have it back again soon ! 

Thanks for your work :)

Palashxeuron:
any updates

Filiflo:
There has been a more recent post on similar issues.
They seem to still be working on it.

here are some references to posts that seem to have the same problem:
[typescript - same issue(last addressed 2 weeks from writing)](https://github.com/google/mediapipe/issues/4605)
[duplicate of this issue - has been marked as such](https://github.com/google/mediapipe/issues/4536#issuecomment-1592460079)

kuaashish:
HI @jays0606,

Could you please verify the current status using the nightly version at https://www.npmjs.com/package/@mediapipe/tasks-vision/v/0.10.12-rc.20240407? It must return the visibility due to the merged https://github.com/google/mediapipe/pull/5142. These changes are expected to be included in the upcoming major release.

Thank you!!

SinchanKM:
It's available for mediapipe web v0.10.11 released on 8th March 2024

PR #5142, opened 2024-02-14
Add visibility field for npm's pose landmarks

lin826:
According to the doc: https://developers.google.com/mediapipe/solutions/vision/pose_landmarker/web_js

This should be supported.

Related to #4409 and #4605


schmidt-sebastian (approved):

lin826:
Just a follow-up: Are there any internal concerns or obstacles to supporting this field?

schmidt-sebastian:
I am working on this but this is breaking a lot of our tests so the change is much bigger than in this PR.

PR #5145, opened 2024-02-15
Fixed iOS CocoaPods conflicts with OpenCV 4.3.0 and TensorFlowLiteSwift CocoaPods

priankakariatyml:


third_party/opencv_ios_source.BUILD
```diff
@@ -42,6 +42,8 @@ genrule(
         --without objdetect \
         --without gapi \
         --without flann \
+        --without highgui \
+        --without videoio \
```
priankakariatyml, line 46: These symbols from these opencv modules conflict with our library when installed with the opencv framework using CocoaPods. I am not sure why the other symbols don't conflict. 
In any case, we don't need these. So I excluded them to avoid the conflicts

schmidt-sebastian (approved):

kivancPaycore:
this is the podfile that I use.
# Uncomment the next line to define a global platform for your project
platform :ios, '16.0'

target 'KYC' do
  # Comment the next line if you don't want to use dynamic frameworks
  use_frameworks!
	
  pod 'MediaPipeTasksVision', '0.10.10'
  pod 'OpenCV'
  # Pods for KYC

end

When i try to import opencv in my .swift file it does not appear in import options. Any suggestions ?

kers97:
this is the podfile that I use.

platform :ios, '16.0'
target 'KYC' do
  use_frameworks!
  pod 'MediaPipeTasksVision', '0.10.10'
  pod 'OpenCV'
end

When i try to import opencv in my .swift file it does not appear in import options. Any suggestions ?

BaronWj:
I also encountered the same problem. Have you solved it?

PR #5175, opened 2024-02-27
Added iOS LLM Task Runner and Inference

priankakariatyml:
1. Added iOS LlmTaskRunner
2. Added iOS LlmInferenceError and LlmInference
3. Updated //mediapipe/tasks/cc/genai/inference/c/llm_inference_engine.h to support C

### DIFF
--- a/mediapipe/tasks/cc/genai/inference/c/llm_inference_engine.h
+++ b/mediapipe/tasks/cc/genai/inference/c/llm_inference_engine.h
@@ -15,8 +15,15 @@
 #ifndef MEDIAPIPE_TASKS_GENAI_INFERENCE_C_LLM_INFERENCE_ENGINE_H_
 #define MEDIAPIPE_TASKS_GENAI_INFERENCE_C_LLM_INFERENCE_ENGINE_H_
 
+#ifdef __cplusplus
 #include <cstddef>
 #include <cstdint>
+
+#else
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+#endif
 
 #ifndef ODML_EXPORT
 #define ODML_EXPORT __attribute__((visibility("default")))
--- a/mediapipe/tasks/ios/genai/inference/BUILD
+++ b/mediapipe/tasks/ios/genai/inference/BUILD
@@ -1,0 +1,59 @@
+# Copyright 2024 The MediaPipe Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
+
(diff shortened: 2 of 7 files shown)


mediapipe/tasks/cc/genai/inference/c/llm_inference_engine.h
```diff
@@ -15,9 +15,16 @@
 #ifndef MEDIAPIPE_TASKS_GENAI_INFERENCE_C_LLM_INFERENCE_ENGINE_H_
 #define MEDIAPIPE_TASKS_GENAI_INFERENCE_C_LLM_INFERENCE_ENGINE_H_
 
+#ifdef __cplusplus
```
priankakariatyml, line 18: The other platforms seem to be using this as a C++ file. This is needed for interoperability with the Swift APIs.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// The random seed for sampling tokens.
+    @objc public var randomSeed: Int = 0
+
+    /// Creates a new instance of `Options` with the modelPath and default values of
+    /// `maxSequenceLength`, `topK``, `temperature` and `randomSeed`.
+    /// This function is only intended to be used from Objective C.
+    ///
+    /// - Parameters:
+    ///   - modelPath: The absolute path to a model asset bundle stored locally on the device.
+    @objc public init(modelPath: String) {
```
priankakariatyml, line 122: A more `Swift`y pattern is to provide an init with default params and declare the variables to be constants that can only be initialised once. But default params are not visible to Objective C. Hence adopting this pattern for a cleaner interface in both languages.

mbrenon (approved):

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -0,0 +1,143 @@
...
+
+  private let llmTaskRunner: LlmTaskRunner
+
+  /// Creates a new instance of `LlmInference` with the given options.
+  ///
+  /// - Parameters:
+  ///   - options: The options of type `LlmInference.Options` to use for configuring the
+  /// `LlmInference`.
+  @objc public init(options: Options) {
+    let modelPath = strdup(options.modelPath)
```
mbrenon, line 34: Question: do we need to strdup here if we free the memory once we exit the method? The same behavior would be achieved by just passing the exiting parameter.

priankakariatyml: I get a Swift warning if I don't do this. Logically yes. That was my original implementation.  Something along the lines of  "memory may be freed if you pass in this to the method." I did this to remove it. So users don't see any warnings when they use the library. But it didn't make sense to me either.

mediapipe/tasks/ios/genai/inference/BUILD
```diff
@@ -0,0 +1,30 @@
...
+
+package(default_visibility = ["//mediapipe/tasks/ios:__subpackages__"])
+
+licenses(["notice"])
+
+swift_library(
+    name = "LlmInference",
+    module_name = "LlmInference",
+    generates_header = 1,
+    generated_header_name = "generated_header/LlmInference-Swift.h",
```
mbrenon, line 25: @priankakariatyml Encountering error "The generated header of a Swift module may not contain directory components".

I'll drop the "generated_header/" prefix. Let me know if this leads to trouble on your side.

priankakariatyml: That's okay. This was for my Objc Test. I am not checking them now. When I do, I'll check what the problem is

PR #5187, opened 2024-03-01
Added iOS Gen AI Inference async method

priankakariatyml:
1. Added iOS Gen AI Inference async method.
2. Added iOS bazel targets and build scripts for Gen AI XCFrameworks


mbrenon (approved):

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -53,25 +52,99 @@ public final class LlmTaskRunner {
...
+      free(inputText)
+    }
+  }
+}
+
+extension LlmTaskRunner {
+  private class func responseStrings(from responseContext: LlmResponseContext) -> [String]? {
     guard let cResponseArray = responseContext.response_array else {
-      throw GenAiInferenceError.invalidResponseError
+      return nil
```
mbrenon, line 136: For my understanding, what's the reason we're not sending an error here?

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -53,25 +52,99 @@ public final class LlmTaskRunner {
...
     }
 
     var responseStrings: [String] = []
-
     for responseIndex in 0..<Int(responseContext.response_count) {
+      /// Throw an error if the response string is `NULL`.
       guard let cResponseString = cResponseArray[responseIndex] else {
-        throw GenAiInferenceError.invalidResponseError
+        return nil
```
mbrenon, line 143: Same here. Also here it doesn't correspond to the comment set above, so should I remove this comment altogether?

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -53,25 +52,99 @@ public final class LlmTaskRunner {
...
+      }
+
+      /// Call completion callback if LLM has generated its last response.
+      if responseContext.done {
+        cCallbackInfo.completion()
+      }
+    }
+  }
+
+  deinit {
```
priankakariatyml, line 102: Deinit here

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -13,22 +13,21 @@
 // limitations under the License.
 
 import Foundation
-import LlmInferenceEngineC
+import MediaPipeTasksGenAIC
 
 /// This class is used to create and call appropriate methods on the C `LlmInferenceEngine_Session`
 /// to initialize, execute and terminate any MediaPipe `LlmInference` task.
 public final class LlmTaskRunner {
-  fileprivate typealias CLlmSession = UnsafeMutableRawPointer
```
mbrenon, line 21: Was this change actually needed?

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -53,25 +52,99 @@ public final class LlmTaskRunner {
...
+      inputText: UnsafeMutablePointer<CChar>?, progress: @escaping (ProgressCallback),
+      completion: @escaping (CompletionCallback)
+    ) {
+      self.inputText = inputText
+      self.progress = progress
+      self.completion = completion
+    }
+
+    deinit {
+      free(inputText)
```
mbrenon, line 128: Suggestion from team: should we use "deallocate" instead? Let's keep that for later, but raising that to your attention :)

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
```diff
@@ -53,25 +52,99 @@ public final class LlmTaskRunner {
...
+}
+
+extension LlmTaskRunner {
+  /// A wrapper class for whose object will be used as the C++ callback context.
+  /// The progress and completion callbacks cannot be invoked without a context.
+  class CallbackInfo {
+    typealias ProgressCallback = (_ partialResult: [String]?, _ error: Error?) -> Void
+    typealias CompletionCallback = () -> Void
+
+    let inputText: UnsafeMutablePointer<CChar>?
```
mbrenon, line 114: This raised several questions to be solved in the (near?) future: 
* shall we let the caller know that CallbackInfo takes ownership of inputText?
* or: do we even need to make inputText a member of callback info? it's not immediately obvious why we need to retain the value after the initial call

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
priankakariatyml: dealloc can't be overridden in Swift. deinit is called just before deallocation. That's the recommended method for cleanup

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
priankakariatyml: if the C APIs are not copying the String before delegating to the background thread, it will get deallocated since its lifetime is only for the duration of this method. The method is asynchronous and returns before the response generation callbacks are invoked. So adding them to the context was my solution for this. We can also make this an instance variable in TaskRunner since we don't allow simultaneous calls. But if in future, we change this behaviour, we won't be able to keep track of inputText's from all calls.

I am okay removing this provided the C APIs copy the string. I didn't see that in the C code.

priankakariatyml: Also the CallbackInfo does not take ownership of the inputText. We allocate a new copy with strdup(). CallbackInfo takes ownership of this copy and deallocates it in deinit

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunner.swift
priankakariatyml: Sorry this is a typo. I meant to make this
`private typealias CLlmSession = UnsafeMutableRawPointer`
private is the lowest access level where it will only be visible to the class declaration and extensions defined within same class. 
`fileprivate` means visible to anything declared in the same file even an entity outside the class.

I have fixed this in my new PR. You can choose to change it before merging.

PR #5191, opened 2024-03-04
Added deletion of cache files and updated interface of iOS LlmTaskRunner and LlmInference

priankakariatyml:
1. Moved caching logic to LlmTaskRunner
2. Added method to interface of LlmInference for deleting all caches.
3. Updated property types in LlmInference.Options
4. Added an extra check to see if provided modelPath is valid. As a result updated LlmInference and LlmTaskRunner inits to be throwing methods.
5. Updated access modifiers of some classes.


priankakariatyml:
@mbrenon Our Google workspace is down. In case you need anything for the release, LMK here or how we can communicate.

PR #5206, opened 2024-03-10
Replace libdc1394-22-dev with libdc1394-dev

boswelja:
libdc1394-22 has been removed, and is superseded by libdc1394. Ref. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=963924

This fixes the inability to run the script on newer systems, as `libdc1394-22` no longer exists.


boswelja:
CLA signed, thanks!

boswelja:
I just realized the PR description wasn't worded super clear. I have reworded it to clarify that the script was previously nonfunctional, but works after this change

boswelja:
Seems like there's no interest in this, I'll close off the PR

Issue #5257, opened 2024-03-24
Problem when using mediapipes from inside a worker loaded as a module


martenrichter:
### Have I written custom code (as opposed to using a stock example script provided in MediaPipe)

None

### OS Platform and Distribution

Web

### MediaPipe Tasks SDK version

master branch

### Task name (e.g. Image classification, Gesture recognition etc.)

Image classification but should apply to all tasks

### Programming Language and version (e.g. C++, Python, Java)

Javascript

### Describe the actual behavior

When loading the code I get an importScript exception

### Describe the expected behaviour

That it also works from workers loaded as module

### Standalone code/steps you may have used to try to get what you need

```shell
It is straightforward to reproduce. Just try to load it from a worker, which is loaded as a module, so something like this:

import { ImageSegmenter, FilesetResolver } from '@mediapipe/tasks-vision'
  async initSegmenter() {
    return ImageSegmenter.createFromOptions(
      await FilesetResolver.forVisionTasks(
        new URL(`../node_modules/@mediapipe/tasks-vision/wasm`, import.meta.url)
          .pathname
      ),
      {
        baseOptions: {
          modelAssetPath: new URL(
            './models/selfie_segmenter_landscape.tflite',
            import.meta.url
          )
        },
        canvas: this.canvas,
        outputCategoryMask: true,
        runningMode: 'VIDEO'
      }
    )
  }
```
If your worker is loaded as a worker, e.g.:
```
AVMediaPipe.workerInternal_ = new Worker(
  new URL('./avmediapipeworker.js', import.meta.url),
  {
    type: 'module'
  }
 )
```
You will get an error saying that importScripts is not supported.
Note, as framework with bundlers such as vite, convert you commonJS module to ESM module, converting the worker to commonJS is not an option.
```


### Other info / Complete Logs

```shell
The origin is here:
https://github.com/google/mediapipe/blob/fde195e4a86298c12c300811b3ae239fd7b6238c/mediapipe/web/graph_runner/run_script_helper.ts.template#L13

I also tried to patch it and send a PR by replacing the first part of the code with:

if (typeof importScripts === 'function') {
     try {
        importScripts(scriptUrl.toString());
    } catch (error) { 
        if (error instanceof TypeError) {
          // importScripts does not work in module workers
          const module = await eval('import')(scriptUrl.toString());
          //@ts-ignore
          self.ModuleFactory = module.ModuleFactory
        } else {
          throw error
        }
    }
  } else {
```
[...]

PR #5278, opened 2024-03-29
Fix using mediapipe from worker loaded as module

martenrichter:
When using the mediapipe in Javascript from a worker, the library does not work if the worker is loaded from a module. This patch set addresses the issues so that it also loads successfully in this environment.

Note that this is a first attempt that is far from being final. Since I had to change the compile options of the typescript compiler, the `import()` statement survives compilation. I think rollup and the commonJS plugin will nevertheless provide commonJS code. On the other hand, I also had to increase the JavaScript specification. It may be a better option not to run the code with the import statement through the TS compiler so as not to break compatibility with older JavaScript.

Furthermore, I had to patch the wasm import code. Adding the fix upstream to the code may be better than patching it here.

### DIFF
--- a/mediapipe/framework/tool/mediapipe_files.bzl
+++ b/mediapipe/framework/tool/mediapipe_files.bzl
@@ -14,10 +14,22 @@ def mediapipe_files(srcs):
 
     for src in srcs:
         archive_name = "com_google_mediapipe_%s" % src.replace("/", "_").replace(".", "_")
-        native.genrule(
-            name = "%s_ln" % archive_name,
-            srcs = ["@%s//file" % archive_name],
-            outs = [src],
-            output_to_bindir = 1,
-            cmd = "ln $< $@",
-        )
+        if src.endswith("_internal.js"):
+          native.genrule(
+              name = "%s_ln" % archive_name,
+              srcs = ["@%s//file" % archive_name],
+              outs = [src],
+              output_to_bindir = 1,
+              cmd = """
+              cp $< $@
+              echo 'else {\n  (globalThis || self || window).ModuleFactory = ModuleFactory;\n  (globalThis || self || window).custom_dbg = function(text) {\n    console.warn.apply(console, arguments);\n  }\n}\n' >> $@
+              """
+          )
+        else: 
+          native.genrule(
+              name = "%s_ln" % archive_name,
+              srcs = ["@%s//file" % archive_name],
+              outs = [src],
+              output_to_bindir = 1,
+              cmd = "ln $< $@",
+          )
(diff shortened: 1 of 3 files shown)


martenrichter:
This is a PR belonging to issue:
https://github.com/google/mediapipe/issues/5257

schmidt-sebastian:
Thank you - will import this manually into the various places.

martenrichter:
This sounds good (as it should be integrated at the place where the Wasm is built). Once it is finished, it would be nice if you posted something here or on the issue. Thanks!

joacopaz:
@schmidt-sebastian did you manage to apply the fix?

farleybarillo:
@schmidt-sebastian any news on the fix? The last commit was Oct 5, 2023. This update will save many lives!

Max-Lovell:
Still having issues with this...

schmidt-sebastian:
... sorry., this will be part of the next release. Promise :)

Max-Lovell:
@schmidt-sebastian Thanks!

For anyone struggling with this, some polyfills/shims at the top of my worker got this working in Vite 5 for me:

```
// These must run before any @mediapipe/tasks-vision code is imported.

// Fix Bug 1: MediaPipe accesses `document` in workers for iOS/Mac detection.
// Provide a Proxy stub that returns OffscreenCanvas for createElement("canvas")
// and returns false for all "in" checks (e.g. "ontouchend" in document).
// Polyfill minimal document for MediaPipe's broken iOS detection in workers
// MediaPipe checks "ontouchend" in document to detect iOS - this prevents the crash
if (typeof document === 'undefined') {
    const handler: ProxyHandler<any> = {
        get(_target, prop) {
            if (prop === 'createElement') {
                return (tag: string) => {
                    if (tag === 'canvas') return new OffscreenCanvas(1, 1);
                    return {};
                };
            }
            return undefined;
        },
        has() { return false; },
    };
    (self as any).document = new Proxy({}, handler);
}

// Fix Bug 2: MediaPipe uses self.import() to dynamically load WASM modules.
// Vite's worker transform doesn't expose this. Bridge it to native import().
if (typeof (self as any).import !== 'function') {
    (self as any).import = async (url: string) => {
        const module = await import(/* @vite-ignore */ url);
        if (module.default) {
            (self as any).ModuleFactory = module.default;
        }
        return module;
    };
}

// Fix Bug 3: vision_wasm_module_internal.js (lines 8369-8374) has a strict-mode scoping bug
// where custom_dbg is declared inside an if-block but referenced outside it.
(self as any).custom_dbg = (...args: unknown[]) => {
    const msg = String(args[0]);
    // Suppress MediaPipe's C++ INFO/WARNING lines that get misrouted to stderr
    if (msg.startsWith('INFO:') || msg.startsWith('W0') || msg.startsWith('I0')) {
        return;
    }
    console.warn(...args);
};
```

PR #5317, opened 2024-04-15
Fixed memory deallocation issues in output images created by MPPImage+Utils

priankakariatyml:
1. Code changes for readability
2. Fixed memory cleanup of output MPImages created with source type UIImage/SampleBuffer for Face Styliser.


mediapipe/tasks/ios/vision/core/utils/sources/MPPImage+Utils.mm
```diff
@@ -515,10 +522,28 @@ - (nullable instancetype)initWithCppImage:(const mediapipe::Image &)image
...
+      // For infinite loops like the camera callback, this results in an increase in memory
+      // footprint over time. To avoid this, the`UIImage` is being created in an @autoreleasepool
+      // block which results in the image being released as soon as the block completes. The MPImage
+      // retains the UIImage, to keep the image alive during its lifetime.
+      // (Until caller keeps a reference to the result retured by the task)
+      //
+      // Reference: `Use Local Autorelease Pool Blocks to Reduce Peak Memory Footprint`
+      // https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmAutoreleasePools.html.
+      MPPImage *mpImage;
+      @autoreleasepool {
```
priankakariatyml, line 540: @schmidt-sebastian Wrapping the inference calls inside loops is something we can leave to the users too in scenarios they need to. Normally this is done from within loops when you observe your apps temporary objects are not getting destroyed. Since this is a feature of UIImages, it's not really a memory leak we create. But a lot Swift devs maybe not be aware of such a thing. We don't really know how the APIs are going to be used. So choosing to do it here just to be safe.

Not needed for the other source types (pixel buffer, sample buffer). Their native create methods don't return autoreleased objects. Thus memory gets deallocated as expected.

schmidt-sebastian (approved):

PR #5404, opened 2024-05-15
doc: star history added to README.md

rezakarbasi:
Star history shows the viewers that the repo is growing fast and it's been supported by a relatively active community


rezakarbasi:
Star history shows the viewers that the repo is growing fast and it's been supported by a relatively active community

PR #5407, opened 2024-05-15
Added more methods to iOS MPPAudioTaskRunner

priankakariatyml:
1. Added audio stream methods to audio task runner.
2. Changed type of sample rate to double for consistency with the native AVAudioFormat.

### DIFF
--- a/mediapipe/tasks/ios/audio/core/sources/MPPAudioTaskRunner.mm
+++ b/mediapipe/tasks/ios/audio/core/sources/MPPAudioTaskRunner.mm
@@ -25,16 +25,19 @@
 
 namespace {
 using ::mediapipe::Packet;
+using ::mediapipe::Timestamp;
 using ::mediapipe::tasks::core::PacketMap;
 using ::mediapipe::tasks::core::PacketsCallback;
 }  // namespace
 
 static NSString *const kTaskPrefix = @"com.mediapipe.tasks.audio";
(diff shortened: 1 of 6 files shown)


mediapipe/tasks/ios/audio/core/sources/MPPAudioTaskRunner.mm
```diff
@@ -127,6 +131,70 @@ - (nullable instancetype)initWithTaskInfo:(MPPTaskInfo *)taskInfo
...
+    [MPPCommonUtils
+        createCustomError:error
+                 withCode:MPPTasksErrorCodeInvalidArgumentError
+              description:[NSString stringWithFormat:@"The audio task is not initialized with "
+                                                     @"audio stream mode. Current Running Mode: %@",
+                                                     MPPAudioRunningModeDisplayName(_runningMode)]];
+    return NO;
+  }
+
+  if (_sampleRate <= 0) {
```
schmidt-sebastian, line 173: Maybe add comment that this means the value hasn't been initialized.

schmidt-sebastian (approved):

mediapipe/tasks/ios/audio/core/sources/MPPAudioTaskRunner.mm
kartynnik: I'd suggest comparing to the default value instead for clarity.

Issue #4608, opened 2023-07-12
async detection in JavaScript


hschwane:
### Have I written custom code (as opposed to using a stock example script provided in MediaPipe)

None

### OS Platform and Distribution

all platforms

### Task name (e.g. Image classification, Gesture recognition etc.)

FaceLandmarker

### Programming Language and version (e.g. C++, Python, Java)

JavaScript

### Describe the actual behavior

```shell
detect() and detectForVideo() run synchronously, blocking the main thread until detection is done.
```

### Describe the expected behaviour

```shell
async versions of detect() and detectForVideo() should exist, as they do (according to the documentation) in the java and python versions of the task. Also, in the older FaceMesh solutions API, a async version (using a callback) was the default.
```

### Standalone code/steps you may have used to try to get what you need

```shell
face landmark task: https://codepen.io/mediapipe-preview/pen/OJBVQJm
old face mesh solution: https://codepen.io/mediapipe/pen/KKgVaPJ
```

### Other info / Complete Logs

```shell
Also it seems that at least when using chrome for Android to access the page, detecting landmarks the new task takes about 70% longer than in the old FaceMesh solution. However I am not an expert at all in profiling web apps, so I might be doing something else wrong as well.
```

kuaashish:
Hello @hschwane,

Could you please elaborate your query with complete details, Also provide the reproducible steps you are following to reproduce the issue from our end. Thank you!!

FishWoWater:
@kuaashish @hschwane 
Can we expect an async typescript version of FaceLandmarker's detection? The new FaceLandmarker API `https://codepen.io/mediapipe-preview/pen/OJBVQJm` seems to run a bit slower than the old facemesh(the face template can't follow the live stream person face as smoothly as it did in facemesh). The reason could be: synchronous detection blocks the render thread, which in turn slowers down the render fps. Another possible reason is the extended pipeline to estimate blendshapes.

hschwane:
Hey, sorry for not getting back here sooner. 
@kuaashish I do not have any specific code or issue to reproduce I could share. Simply profiling the examples / codepens for old an new FaceLandmarker shows the difference in performance. And the lack of an async detection function can be seen in the API documentation.

@FishWoWater Yes that is exactly what I observed and what I was wondering as well. Thanks for specifying.

PR #5525, opened 2024-07-10
Modify blendshape graph working based on output_blendshapes setting

col-fi:
We noticed that even when `outputFaceBlendshapes` in `FaceLandmarkerOptions` was set to `false`, the blendshape graph was still running. This update makes sure that the `FaceLandmarkerGraph` doesn't include the blendshape graph when `outputFaceBlendshapes` is `false`.

The performance metrics we checked are as follows:
**Devices**
PocoF1
|  outputFaceBlendshapes   |  True          | False      |False(remove Graph)       |
|-----------------|-------------|-------------|-------------|
| inferenceTime(ms)       | 5554~6280ms      | 163~173ms    |109~114ms    |
| CPU(%)        | 32~38     | 27~38     | 21~25    |

S22
|  outputFaceBlendshapes | True          |  False      | False(remove Graph)       |
|-----------------|-------------|-------------|-------------|
| inferenceTime(ms)       | 2700~2900ms   | 96~103ms    |73~85ms   |
| CPU(%)        | 40~45     | 40~48     | 22~25    |

iPhone7
|  outputFaceBlendshapes   | True          |  False      | False(remove Graph)       |
|-----------------|-------------|-------------|-------------|
| inferenceTime(ms)       | 62.45-131.39ms   | 58.50~128.16ms   |44.6-88.3ms   |
| CPU(%)        | 140%     | 135~138%     | 138%    |

### DIFF
--- a/mediapipe/tasks/cc/vision/face_landmarker/face_landmarker_graph.cc
+++ b/mediapipe/tasks/cc/vision/face_landmarker/face_landmarker_graph.cc
@@ -366,6 +366,14 @@ class FaceLandmarkerGraph : public core::ModelTaskGraph {
               .face_landmarks_detector_graph_options()
               .has_face_blendshapes_graph_options()));
     }
+    // If the has_face_blendshapes_graph_options() value in FaceLandmarksDetectorGraphOptions is true,
+    // the blendshape graph is created. However, FaceLandmarkerGraphOptions includes face_blendshapes_graph_options
+    // by default, which results in the creation of the blendshape graph regardless.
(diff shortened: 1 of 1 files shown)


Theo-Y:
Because of this issue, Web TaskVision is also slower than previous versions in face landmarks. Since the javascript version synchronizes with the graph operation, setting blendshape to false in the output does not make a difference in the computation speed(#4608  ).
Could you please review this PR? @kuaashish

Jeremy293 (approved):

PR #5544, opened 2024-07-23
Update for returning gesture index in gesture_recognizer.py

Sandeepsabbella:
The gesture recognizer is returning a default value of -1. changed the _GESTURE_DEFAULT_INDEX to gesture.index to return the gesture class index (id:int). Now the respective class index is returned when the category is predicted by the model.


Sandeepsabbella (commented):
The code was tested to ensure it returns gesture index
file changes: gesture_recognizer.py

PR #5566, opened 2024-08-12
Added iOS refactored LlmInference, LlmTaskRunner and LlmSessionRunner

priankakariatyml:
1. Added LlmSessionRunner that manages a C session created by a C engine. Each swift Llm session will be managed by its own LlmSessionRunner.
2. Refactored LlmTaskRunner to manage the C Llm Engine. All C methods that require the engine instance are invoked via the LlmTaskRunner
3. Added refactored interface for LlmInference.


mediapipe/tasks/ios/genai/core/sources/LlmTaskRunnerRefactored.swift
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import MediaPipeTasksGenAIC
+
+/// This class is used to create and call appropriate methods on the C `LlmInferenceEngine` to
+/// initialize an LLMInference task and create prompt sessions.
+final class LlmTaskRunner {
```
priankakariatyml, line 20: @yishuangP Can we rename this to `LlmEngineRunner` to keep it closer to the C APIs? Not sure how Java will be structured.

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunnerRefactored.swift
```diff
@@ -0,0 +1,88 @@
...
+      }) == StatusCode.success.rawValue
+    else {
+      throw GenAiInferenceError.failedToInitializeEngine(
+        String(allocatedCErrorMessage: cErrorMessage))
+    }
+  }
+
+  /// Creates a new C LLM session from the current C engine and returns an `LlmSessionRunner` 
+  /// that wraps around the newly created C session. The session runner is responsible for managing 
+  /// its underlying C session.
```
yishuangP, line 45: Shall we mention that this method will always return a new instance even with the same `LlmSessionConfig`?

mediapipe/tasks/ios/genai/core/sources/LlmSessionRunner.swift
```diff
@@ -0,0 +1,209 @@
...
+    let callbackContext = UnsafeMutableRawPointer(Unmanaged.passRetained(callbackInfo).toOpaque())
+
+    LlmInferenceEngine_Session_PredictAsync(cLlmSession, callbackContext) {
+      context, responseContext in
+      guard let cContext = context else {
+        return
+      }
+      guard let cResponse = responseContext?.pointee else {
+        return
+      }
```
yishuangP, line 102: Is this expected? Shall we throw an error here?

mediapipe/tasks/ios/genai/inference/sources/LlmInferenceRefactored.swift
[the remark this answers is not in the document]
priankakariatyml: Done. FYI, until we merge the all the code master will have an incomplete implementation.

mediapipe/tasks/ios/genai/core/sources/LlmTaskRunnerRefactored.swift
priankakariatyml: We should. I have added a note in the comments.

yishuangP (changes requested):
Thanks Prianka, left some comments

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -100,123 +81,23 @@ import MediaPipeTasksGenAIC
...
-      throw GenAiInferenceError.invalidResponse
-    }
-
-    return humanReadableLlmResponse
+  ///   - sessionConfig: The C config of type `LlmSessionConfig` that configures how to execute the 
+  /// model.
+  /// - Returns:
+  ///   - An `LlmSessionRunner` that wraps around a new session.
+  /// - Throws: An error if the underlying engine could not create a session.
+  func createSessionRunner(sessionConfig: LlmSessionConfig) throws -> LlmSessionRunner {
```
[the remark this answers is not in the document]
priankakariatyml: We do have Session Options in swift. I have not committed the file for the `Session` to keep the current PR size manageable. Swift Options are also declared in that file. 
This method is not public. It has internal visibility and is called by the `init` of the swift `Session`. `Session.init(llmInference: options:)` creates the C struct `LlmSessionConfig` from the swift options provided by the user and then calls `llmInference.createSessionRunner(sessionConfig)`.  

I'll commit the rest of the code once this PR is merged.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -21,70 +21,51 @@ import MediaPipeTasksGenAIC
...
+
+    /// Copying to dynamically allocated property on heap ensures that the `supportedLoraRanks` will
+    /// be in memory until this `LLmInference` is deallocated. `supportedLoraRanks` is deallocated
+    /// in `deinit`.
+    ///
+    /// Assigning the pointer from `withUnsafeMutablePointer()` to a property does not guarantee the
+    /// array being in memory as long as the engine isn't deallocated since
+    /// `options.supportedLoraRanks` only has function scope.
+    ///
+    /// TODO: If C++ API mem copies the array the following code can be updated to
```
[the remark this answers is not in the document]
priankakariatyml: I have updated to `withUnsfeMutable`. Also removed allocation of modelPath and cacheDir. They can also be passed using `withCString`.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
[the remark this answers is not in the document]
priankakariatyml: We can also change this method to `createSessionRunner(options: Session.Options)` and let this method handle creation of the C session config from the options passed in by the Session. I just thought it would be better for the Session to convert its options to the config however it sees fit and not let the LlmInference worry about it.
LMK what you think.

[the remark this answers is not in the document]
yishuangP: > This method is not public. It has internal visibility and is called by the init of the swift Session. Session.init(llmInference: options:) creates the C struct LlmSessionConfig from the swift options provided by the user and then calls llmInference.createSessionRunner(sessionConfig).

Thanks for the explanation!  If you don't mind, can we make all changes in one PR? We have some internal dependency on this swift API, I hope we can make all necessary changes in one PR. Let me know if you have other opinions.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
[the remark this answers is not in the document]
yishuangP: Sorry it seems that you haven't committed the changes?

yishuangP (changes requested):

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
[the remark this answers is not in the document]
priankakariatyml: I have pushed all the code to this PR. Do have a look

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -21,70 +21,34 @@ import MediaPipeTasksGenAIC
...
-      free(loraPath)
-    }
-    var loraRanks: UnsafeMutableBufferPointer<Int>?
-    options.supportedLoraRanks.withUnsafeMutableBufferPointer { pointer in
-      loraRanks = pointer
+    let cacheDirectory = FileManager.default.temporaryDirectory.path
+
+    llmTaskRunner = try options.modelPath.withCString { modelPath in
+      try cacheDirectory.withCString { cacheDirectory in
+        try options.supportedLoraRanks.withUnsafeMutableBufferPointer { supportedLoraRanks in
```
priankakariatyml, line 38: w.r.t https://github.com/google-ai-edge/mediapipe/pull/5566#discussion_r1723766031
@yishuangP These changes were already pushed. I think there was some git refresh issue.

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
```diff
@@ -0,0 +1,289 @@
...
+        responseGenerationInProgressQueue.async(flags: .barrier) {
+          self.responseGenerationInProgressInternal = newValue
+        }
+      }
+    }
+
+    /// Creates a new instance of `LlmInference` with the given options. 
+    /// Ensure that the instance of `LlmInference` passed to this function must be alive through 
+    /// the lifetime of the current session. Using a session beyond the life time of the 
+    /// `LlmInference` used to create it will lead to undefined behaviour.
```
priankakariatyml, line 59: @yishuangP I decided to not store the llmInference in the session since it is not used in any API other than creation of the Session.  But to avoid the llmInference getting deallocated before all the sessions created from it are destroyed, we can store it in the session as an instance variable. Won't lead to a retain cycle since `LlmInference` does not have any reference to `Session`. Just that the `llmInference` may remain in memory even if it goes out of scope in the user program until all sessions are destroyed. LMK your thoughts. We can do this in a follow up PR if needed. Not critical at the moment.

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
```diff
@@ -0,0 +1,269 @@
...
+  /// the same time.
+  /// You can also clone an existing session and continue querying the LLM from where you left off.
+  ///
+  /// Note: Inherits from `NSObject` for Objective C interoperability.
+  @objc(MPPLLMInferenceSession) public final class Session: NSObject {
+    // Session runner that manages the creation, deletion and execution of the underlying C session.
+    private let llmSessionRunner: LlmSessionRunner
+
+    // LLM Inference used to create this session.
+    private let llmInference: LlmInference
```
priankakariatyml, line 32: I am storing a strong reference to `llmInference` to ensure that any given session does not out live the `llmInference` and create all sorts of undefined behaviour. This won't cause a retain cycle since llmInference does not store a reference to any session.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -100,123 +80,49 @@ import MediaPipeTasksGenAIC
...
-            continuation.finish()
-          })
-      } catch {
-        continuation.finish(throwing: error)
+  /// If no response generation using this `llmInference` is currently in progress, this function
+  /// updates the response generation state and returns successfully granting access to its caller
+  /// to execute response generation. If response generation is already in progress, throws an
+  /// error.
+  /// Any
+  func shouldContinueWithResponseGeneration() throws {
```
priankakariatyml, line 102: @yishuangP I am maintaining a response generation state in the inference which the session can read and update through these methods to prevent simultaneous response generation calls from multiple sessions.

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
yishuangP: Thanks! Yeah agree that we need a strong reference to `llmInference` here to manage life cycles. 

> Also lets me synchronise the response generation calls across sessions to fix a crash because of simultaneous response generation calls.

I believe only 1 response generation can happen at a time, is this what you refer to?

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
```diff
@@ -0,0 +1,291 @@
...
+      /// Used to make a decision about whitespace stripping.
+      var receivedFirstToken = true
+
+      try llmSessionRunner.predictAsync(
+        progress: { partialResponseStrings, error in
+          guard let responseStrings = partialResponseStrings,
+            let humanReadableLlmResponse = Session.humanReadableString(
+              llmResponses: responseStrings, stripLeadingWhitespaces: receivedFirstToken)
+          else {
+            progress(nil, GenAiInferenceError.invalidResponse)
```
yishuangP, line 168: We should also call `self?.llmInference.markResponseGenerationCompleted()` if it errors out?

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -17,6 +17,10 @@ import MediaPipeTasksGenAIC
 
 /// A MediaPipe task that performs inference using a given Large Language Model.
 ///
+/// An instance of `LlmInference` will only be deallocated after all sessions created from it are
+/// destroyed. This means that an LLM inference can stay in memory even if a reference to it goes
+/// out of scope if atleast one of its session's outlives its scope.
```
yishuangP, line 22: super nit: out of scope if at least one of its sessions outlives its scope.

mediapipe/tasks/ios/genai/inference/sources/LlmInference.swift
```diff
@@ -26,71 +30,52 @@ import MediaPipeTasksGenAIC
...
-      }
-    }
-  }
-  private var supportedLoraRanks: UnsafeMutableBufferPointer<Int>?
+  private var responseGenerationInProgress = false
 
   /// Creates a new instance of `LlmInference` with the given options.
+  /// An instance of `LlmInference` will only be deallocated after all sessions created from it are
+  /// destroyed. This means that an LLM inference can stay in memory even if the reference to it
+  /// goes out of scope if atleast one of its session's outlives its scope.
```
yishuangP, line 45: super nit: out of scope if at least one of its sessions outlives its scope.

yishuangP (changes requested):
Thanks a lot! Can you help fix the sample app examples/llm_inference/ios/InferenceExample in a follow-up PR?

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
priankakariatyml: yes. across sessions.

mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
priankakariatyml: You mean if it errors out in the progress callback? I am assuming, even if one of the partial responses is invalid, the C++ session would still continue giving callbacks for the remaining responses? 
In that case wouldn't marking it completed, still cause the same issue? By design, irrespective of whether the response returned by C++ is invalid, the session runner will also call `completion()` after `progress()` if the response context marks the bool `done` as true.

LMK if this is not how C++ behaves.

priankakariatyml: But thanks to your query, I spotted an issue in the control flow of sync `predict()` when an error happens and have fixed the same.

yishuangP: Thanks for the explanation. I took a look at llmSessionRunner.predictAsync implementation, we don't need to call `self?.llmInference.markResponseGenerationCompleted()` here.

yishuangP (approved):
Thanks!

yishuangP (approved):

PR #5635, opened 2024-09-19
Fixed leading white space stripping and added new options to iOS LlmInference

priankakariatyml:
1. Fixed leading whitespace stripping by stripping only leading white spaces from first non empty response.
2. Added new options

### DIFF
--- a/mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
+++ b/mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
@@ -171,8 +171,14 @@ extension LlmInference {
             return
           }
 
-          /// Reset state after first response is processed.
-          receivedFirstToken = false
+          /// Reset state after first non empty response is processed. Ensures that leading 
+          /// whitespaces are stripped from the first non empty response.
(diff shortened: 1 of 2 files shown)


mediapipe/tasks/ios/genai/inference/sources/LlmInference+Session.swift
```diff
@@ -159,20 +159,26 @@ extension LlmInference {
...
-      var receivedFirstToken = true
+      var receivedFirstNonEmptyToken = false
 
       llmSessionRunner.predictAsync(
         progress: { partialResponseStrings, error in
           guard let responseStrings = partialResponseStrings,
             let humanReadableLlmResponse = Session.humanReadableString(
-              llmResponses: responseStrings, stripLeadingWhitespaces: receivedFirstToken)
+              llmResponses: responseStrings, stripLeadingWhitespaces: !receivedFirstNonEmptyToken)
           else {
```
priankakariatyml, line 169: @yishuangP I have changed the name of the state variable and the values it takes in different scenarios to map closely to the  meaning of the  variable name for better readability. So the condition check for leading white space stripping is flipped. The logic remains the same as the previous implementation.

yishuangP (approved):

PR #5723, opened 2024-11-11
Update requirements.txt to allow Numpy 2.0.2

haiggoh:
allow Numpy 2.0.2 since 1.26.4 is becoming increasingly outdated leading to dependency conflicts with newer versions of pytorch.


lenoch0d:
Could we merge this?

PR #5939, opened 2025-04-09
Fix typing

Ioann44:
Fix warning using `mp_pose.POSE_CONNECTIONS` (frozenset) in `draw_landmarks` function

### DIFF
--- a/mediapipe/python/solutions/drawing_utils.py
+++ b/mediapipe/python/solutions/drawing_utils.py
@@ -15,7 +15,7 @@
 
 import dataclasses
 import math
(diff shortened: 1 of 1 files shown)


Vessel9817:
Could be even more abstract and use `Sequence` or `Iterable`, given that iterating with `for connection in connections:` is the only use of `connections`.

Ioann44:
I guess there was a reason to try `Container` in my first PR, so `Iterable` not suitable. Also, `Sequence` is less abstact, as it require indexed access.
@schmidt-sebastian

PR #5941, opened 2025-04-14
Tests working

Ananay-22:
Preliminary tests on the landmarkers seem to show that they are working. Need to propogate and figure out how to run all tests


Ananay-22:
Hi, this is my first PR here I think - please let me know the protocol if I missed it. 

https://github.com/google-ai-edge/mediapipe/issues/1620

This is the relevant issue - you can track my changes with the comments here.

Heads up - the issue was marked closed but I request it be re-opened since the underlying problem was not resolved (requested landmark confidence but solution is handedness confidence)

PR #5942, opened 2025-04-14
Fixed python hand palm connection

quagrain:
I noticed the palm landmark connections in the documentation did not correlate with the Python implementation. It's implemented as (1,5) in python but the docs show it should be (0,5). I also confirmed with the `cc`, `ios`, `java` and `web` implementations that confirm that it should be (0,5).


kuaashish:
Hi @schmidt-sebastian,

Could you please have a look into this issue? 

Thank you!!

schmidt-sebastian (approved):

PR #5949, opened 2025-04-23
Enable multi-class NMS for object detector

xKHUNx:
This pull request enabled multi-class NMS for object detectors in vision package, for all APIs including iOS, Java, Python and web.

Also included in this pull request:
- Fixed Dockerfile.manylinux_2_28_x86_64 file for building Python wheel.
- Added script to build tasks for web API with script modified from https://github.com/whatisor/mediapipe/blob/master/build_tasks_vision.sh.


xKHUNx:
@schmidt-sebastian Hey, can you took a look at this PR. I believe this is a useful feature to be enabled, since it's already implemented but not enabled for the other platforms.

PR #6076, opened 2025-09-03
Import numpy as np missing in holistic docs

heymuun:
Problem: missing import numpy as np in holistic.md
Update: I have signed up as cla


PareekHeymun:
Merge it, otherwise I will steal your dog.

PR #6186, opened 2025-12-16
Python: Enabling numpy >=2.0 and up to Python 3.13

MSDNAndi:
Enabling numpy >=2.0 and Python 3.13, also upgraded python tensorflow dependencies in model-maker


MSDNAndi:
It seems mediapipe 0.10.30 with numpy 2.0 support just showed up on https://pypi.org/project/mediapipe/ - so my PR proposal is obsolete

Issue #5283, opened 2024-04-01
ImageProcessingOptions package is not correct in docs


JC3:
### Description of issue (what needs changing)

Whatever package `ImageProcessingOptions` is in is not documented correctly.

### Clear description

https://developers.google.com/mediapipe/api/solutions/python/mp/tasks/vision/holistic_landmarker/image_processing_options_module/ImageProcessingOptions shows that `ImageProcessingOptions` is in `mediapipe.tasks.vision.holistic_landmarker.image_processing_options_module` but this package does not exist. I have no idea where this class is exported from.

kinarr:
Adding the following code to https://github.com/google/mediapipe/blob/master/mediapipe/tasks/python/vision/__init__.py should fix it most likely:

```python
ImageProcessingOptions = core.image_processing_options.ImageProcessingOptions

del holistic_landmarker
```

markmcd:
> https://developers.google.com/mediapipe/api/solutions/python/mp/tasks/vision/holistic_landmarker/image_processing_options_module/ImageProcessingOptions shows that `ImageProcessingOptions` is in `mediapipe.tasks.vision.holistic_landmarker.image_processing_options_module` but this package does not exist. I have no idea where this class is exported from.

If you take the fully-qualified class from that URL, it exists:

```py
$ pip install mediapipe
...
$ python
>>> import mediapipe as mp
>>> mp.__version__
'0.10.11'

>>> mp.tasks.vision.holistic_landmarker.image_processing_options_module.ImageProcessingOptions()
ImageProcessingOptions(region_of_interest=None, rotation_degrees=0)

>>> mp.tasks.vision.holistic_landmarker.image_processing_options_module
<module 'mediapipe.tasks.python.vision.core.image_processing_options' from '/tmp/mp/lib/python3.11/site-packages/mediapipe/tasks/python/vision/core/image_processing_options.py'>
```

> I found it in `mediapipe.tasks.python.vision.core.image_processing_options`

This looks like the path from the source directory, not the `mediapipe` pip module. i.e. if you replace the `.`s with `/`, you can find it on GitHub [here](https://github.com/google/mediapipe/blob/master/mediapipe/tasks/python/vision/core/image_processing_options.py).


> Also `mediapipe.tasks.components.containers` is another one that doesn't exist, it's actually `mediapipe.tasks.python.components.containers`.

This is the same. If you use the namespace from the docs directly in Python, you'll get the module you expect.

[...]

PR #6273, opened 2026-04-12
docs: expose missing NormalizedKeypoint and cosine_similarity exports

satwikachanta-glitch:
This PR addresses parts of Issue #5283.

I have exposed NormalizedKeypoint in components.containers and cosine_similarity in components.utils. These exports were missing from the respective __init__.py files, which caused a mismatch between the official documentation paths and the actual Python API availability.


satwikachanta-glitch:
cc @markmcd @kuaashish — Whenever you have a chance, could you check and approve the workflow runs so the automated checks can complete?

Issue #6276, opened 2026-04-22
SIGSEGV in AndroidPacketCreator.nativeCreateRgbaImage called from ModelResourcesCacheService.nativeInstallServiceObject


deputat007:
### Have I written custom code (as opposed to using a stock example script provided in MediaPipe)

Yes

### OS Platform and Distribution

Android 8.1.0, 13, 15, 16 (see affected devices table)

### Mobile device if the issue happens on mobile device

| Device | OS Version | Chipset Family |
| :--- | :--- | :--- |
| **Motorola Moto G84 5G** | Android 15 | Qualcomm Snapdragon 695 (arm64-v8a) |
| **Samsung Galaxy M55 5G** | Android 13 | Samsung (arm64-v8a) |
| **Samsung Galaxy S20 FE 5G** | Android 8.1.0 | Qualcomm (arm64-v8a) |
| **Samsung Galaxy Tab A (2016) With S Pen** | Android 16 | (arm64-v8a / armeabi-v7a) |
| **Samsung Galaxy S21 FE 5G** | Android 16 | Qualcomm (arm64-v8a) |
| **Samsung Galaxy S25** | Android 16 | Qualcomm Snapdragon 8 Elite (arm64-v8a) |

### Programming Language and version

Kotlin 2.3.10 / Java 17

### MediaPipe version

0.10.29

### Solution

GestureRecognizer + FaceDetector (Tasks Vision, RunningMode.IMAGE)

### Android Studio, NDK, SDK versions (if issue is related to building in Android environment)

AGP 8.13.2 · compileSdk/targetSdk 35 · minSdk 24 · LiteRT 1.4.0

### Describe the actual behavior

The app crashes with SIGSEGV inside libmediapipe_tasks_vision_jni.so during task-creation warmup (not during inference). The crash is reproducible across 6 different devices, 4 Android OS versions (8.1 through 16)

### Describe the expected behaviour

createFromOptions() should complete without crashing. The AndroidPacketCreator native handle must be valid before nativeInstallServiceObject invokes nativeCreateRgbaImage during the graph warmup sequence.

### Standalone code/steps you may have used to try to get what you need

The crash occurs in a flow that: 

1. Creates a GestureRecognizer via GestureRecognizer.createFromOptions(context, options) on a background thread (RunningMode.IMAGE, Delegate.CPU). 

[...]

PR #6277, opened 2026-04-23
genai converter: reject path-traversal in safetensors tensor names

cybrdude:
### What

Fixes a path-traversal issue in the MediaPipe GenAI safetensors
converter at
`mediapipe/tasks/python/genai/converter/weight_bins_writer.py`.

Tensor names parsed from a `.safetensors` file's JSON header flow into
four `os.path.join(self._output_dir, var_name)` call sites
(lines 96, 100, 103, 114) and are written to disk via
`f.write(output.tobytes())`. No character-class check, no anchoring to
`output_dir`. A tensor named with traversal tokens (`../`), an
absolute path (`/tmp/...`), or embedded path separators causes the
writer to create files outside the declared output directory with the
attacker's tensor bytes.

The call chain is:

1. `safetensors_converter._SafetensorsShardReader.__init__` reads the
   header and runs it through `json.loads`; tensor names become dict
   keys.
2. `GemmaMapper.update_target_name` (the only transformation applied
   to names) performs `.replace()` calls for known Gemma layer names
   and does not touch `/` or `../`.
3. The untouched name reaches the writer sites above.

### How

Added a module-level `_safe_join(output_dir, var_name)` helper that:

1. Rejects any `var_name` not matching `^[A-Za-z0-9._-]+$`.
2. Anchors the resolved path to `realpath(output_dir)` via
   `os.path.commonpath`, rejecting anything that escapes.

Routed all four call sites in `WeightBinsWriter.write_variables()`
through it.

### Test

`mediapipe/tasks/python/genai/converter/tests/weight_bins_writer_path_safety_test.py`
covers:

- Relative traversal rejected (`../escape.bin`)
- Absolute path rejected (`/tmp/escape.bin`)
- Deep traversal rejected (`../../../../../tmp/escape.bin`)
- Embedded separator rejected (`subdir/file.bin`)
- Safe names accepted (`layer0.w`, `emb-table_1.bin`, `alpha-beta`,
  `a.b.c`)

To run:

```bash
python -m pytest mediapipe/tasks/python/genai/converter/tests/\
weight_bins_writer_path_safety_test.py
```

### DIFF
--- a/mediapipe/tasks/python/genai/converter/tests/weight_bins_writer_path_safety_test.py
+++ b/mediapipe/tasks/python/genai/converter/tests/weight_bins_writer_path_safety_test.py
@@ -1,0 +1,116 @@
+# Copyright 2026 The MediaPipe Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+
+"""Regression tests for path-traversal hardening in weight_bins_writer.
+
+Exercises the real weight_bins_writer._safe_join. Before importing the
+module, we register placeholder packages in sys.modules so that
+absolute imports inside weight_bins_writer resolve to our stubs
+instead of triggering mediapipe.tasks.python.__init__, which eagerly
+imports cv2, matplotlib, and jax for unrelated vision utilities.
+"""
+
+import importlib.util
+import os
+import pathlib
+import sys
+import tempfile
+import types
+import unittest
+
+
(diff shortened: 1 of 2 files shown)


cybrdude:
This PR Closes https://github.com/google-ai-edge/mediapipe/issues/6278

cybrdude:
@ahamedjobayer081-spec 

Issue #6276 is a completely unrelated Android SIGSEGV report from a different user (deputat007), filed April 22. It has nothing to do with the path-traversal fix I submitted.  

Why did you add the comment?

schmidt-sebastian:
Closing this as we are removing GenAI Converter from MediaPipe.

PR #6334, opened 2026-08-14
Bounds-check zip entry position/size in ExtractFilesfromZipFile

hdk10:
### Summary
`ExtractFilesfromZipFile` builds an `absl::string_view` directly into the model buffer using
`zip_file_info.position` and `zip_file_info.size`, both of which come from the zip central
directory (`uncompressed_size` / stream position) and are never validated against `buffer_size`.
A crafted `.task`/model bundle can therefore point an entry past the end of the buffer, producing a
heap out-of-bounds read when the referenced bytes are later consumed.

This adds an overflow-safe bounds check before constructing the `string_view`.

### Impact
Out-of-bounds read (crash / potential info disclosure) when loading a crafted model bundle via the
Tasks API (e.g. `BaseOptions(model_asset_path=...)`), before inference.

### Change
Reject any entry where `position > buffer_size` or `size > buffer_size - position`. No behavior
change for valid archives.

### Notes
Reported through the Google OSS VRP (issue 546472889). Verified with an AddressSanitizer PoC against
this code path. Analysis was AI-assisted and human-reviewed; I was not able to run the full Bazel
build locally, so please run CI.

### DIFF
--- a/mediapipe/tasks/cc/metadata/utils/zip_utils.cc
+++ b/mediapipe/tasks/cc/metadata/utils/zip_utils.cc
@@ -142,6 +142,18 @@ absl::Status ExtractFilesfromZipFile(
     int error = unzGoToFirstFile(zf);
     while (error == UNZ_OK) {
       ABSL_ASSIGN_OR_RETURN(auto zip_file_info, GetCurrentZipFileInfo(zf));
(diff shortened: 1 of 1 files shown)


hdk10:
@googlebot I signed it!

PR #6336, opened 2026-08-14
Verify untrusted embedded flatbuffer/flexbuffer before parsing in detection_postprocessing_graph

hdk10:
### Summary
`detection_postprocessing_graph.cc` parses two untrusted, model-embedded buffers with no verification,
plus one missing null-check:

1. **Nested `ObjectDetectorOptions` flatbuffer.** `ConfigureOutModelNmsTensorsToDetectionsCalculator`
   and `ConfigureSsdAnchorsCalculator` call `GetObjectDetectorOptions(custom_metadata->data()->data())`
   (a bare `GetRoot`, no `Verifier`) on the attacker-controlled `DETECTOR_METADATA` `CustomMetadata`
   blob. The outer metadata verifier only checks the byte-vector length, not that the bytes form a
   valid nested flatbuffer, so a malformed blob leads to an out-of-bounds read / wild-pointer walk.

2. **`flexbuffers::GetRoot` on `custom_options`.** `GetMaxClassesPerDetection` calls
   `flexbuffers::GetRoot(custom_options->Data(), custom_options->size())` with no
   `flexbuffers::VerifyBuffer`. `GetRoot` reads the trailing bytes to seed parsing, so a 0/1-byte
   `custom_options` underflows before the buffer start.

3. **Null `custom_code()`.** The op-code search dereferences `custom_code()->str()` without checking
   `custom_code() != nullptr` (an optional field), crashing on a CUSTOM op-code lacking it.

### Impact
Out-of-bounds read / null dereference (crash / potential info disclosure) during
`ObjectDetector::Create()` on a crafted model, before inference.

### Change
Verify the `ObjectDetectorOptions` buffer with `VerifyObjectDetectorOptionsBuffer` before parsing (both
sites); verify the flexbuffer with `flexbuffers::VerifyBuffer` before `GetRoot`; null-check
`custom_code()`. No behavior change for valid models.

### Notes
Reported through the Google OSS VRP (issue 546472889). Verified with AddressSanitizer PoCs against the
`GetObjectDetectorOptions` and `flexbuffers::GetRoot` sinks. Analysis was AI-assisted and
human-reviewed; I could not run the full Bazel build locally, so please run CI.

### DIFF
--- a/mediapipe/tasks/cc/components/processors/detection_postprocessing_graph.cc
+++ b/mediapipe/tasks/cc/components/processors/detection_postprocessing_graph.cc
@@ -397,6 +397,7 @@ int GetMaxClassesPerDetection(const tflite::Model& model) {
       model.operator_codes()->begin(), model.operator_codes()->end(),
       [](const auto& op_code) {
         return op_code->builtin_code() == tflite::BuiltinOperator_CUSTOM &&
+               op_code->custom_code() != nullptr &&
                op_code->custom_code()->str() == kDetectionPostProcessOpName;
       });
   if (op_code_it == model.operator_codes()->end()) {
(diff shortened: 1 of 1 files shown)


hdk10:
@googlebot I signed it!

Issue #6294, opened 2026-05-11
[Feature Request] Extract supported labels from model at runtime


marcusx2:
I have a requirement for an application where the user can change the model of the object detector at runtime, and he should be able to see the supported labels of that model at runtime. For example, [this model](https://storage.googleapis.com/mediapipe-models/object_detector/efficientdet_lite0/float32/latest/efficientdet_lite0.tflite). I can work around this for the default models because the labels are known, but the user can upload his own model. Can a method be added to display the supported labels of a given model? This is for Mediapipe Solutions Web.

marcusx2:
[schmidt-sebastian](https://github.com/schmidt-sebastian) Are there plans to support this?

a2105z:
I'll take this.

@HarishPermal @schmidt-sebastian @kuaashish @tyrmullen @kalyan2789g @akashvverma1995 — claiming #6294.

**What the reporter needs:** an app that lets users swap/upload a custom object-detector `.tflite` cannot show that model's class list unless the labels are hardcoded. `detect()` only returns categories for objects that actually fire, so unused classes stay invisible.

**Fix:** add `ObjectDetector.getLabels(): string[]` on the Web Tasks API, matching `ImageSegmenter.getLabels()`. Labels are already in TFLite Model Metadata and get written onto `DetectionLabelIdToTextCalculator` when the graph is built. I'll read that labelmap after graph refresh and return `[]` when a model has no label file.

PR incoming.

a2105z:
PR is up: https://github.com/google-ai-edge/mediapipe/pull/6339

@HarishPermal @schmidt-sebastian @kuaashish — ready for review. Usage after loading a (possibly user-uploaded) model:

```js
const detector = await ObjectDetector.createFromOptions(vision, {
  baseOptions: { modelAssetPath: userModelUrl },
});
const labels = detector.getLabels(); // e.g. ["person", "bicycle", "car", ...]
```

No inference required. Empty array if the `.tflite` has no labelmap in metadata.

marcusx2:
[...]

PR #6347, opened 2026-08-21
Add getLabels() and getDisplayNames() on Web classifiers

a2105z:
## Summary

Follow-up to #6339 / #6294. `@marcusx2` asked for locale `displayNames` (not just `categoryName`) and for the same runtime label API on every task that returns `Category`, not only ObjectDetector.

- `ImageClassifier`, `TextClassifier`, and `AudioClassifier` now expose `getLabels(): string[]` and `getDisplayNames(): string[]`
- `ImageSegmenter` already had `getLabels()`; it now also has `getDisplayNames()`
- If the model has a display-name associated file, `getDisplayNames()` is the same length as `getLabels()` and index-aligned. If it has none, `getDisplayNames()` returns `[]` (same “absent” signal as `getLabels()`). No inference required.
- Shared helper reads the TFLite labelmap already written onto `TensorsToClassificationCalculator` / `TensorsToSegmentationCalculator` (same pattern as #6339 / `ImageSegmenter.getLabels()`)

ObjectDetector `getDisplayNames()` is intentionally not in this PR so it does not conflict with #6339. Once that lands it is a small add on the same calculator (`LabelMapItem.display_name`).

```ts
const classifier = await ImageClassifier.createFromOptions(vision, {
  baseOptions: { modelAssetPath: userModelUrl },
  displayNamesLocale: 'es',
});
classifier.getLabels();        // ["apple", "banana"]
classifier.getDisplayNames();  // ["manzana", "plátano"]  or  []
```

## Test plan

- Unit tests for empty labelmap, labels+display names, and labels without display names on ImageClassifier
- Same coverage on TextClassifier, AudioClassifier, ImageSegmenter
- Helper tests for index alignment and the unique-calculator guard
- Manual: load a custom classifier `.tflite` with Associated Files / display names and call both getters before `classify()`

Related to #6294

@marcusx2 @HarishPermal @schmidt-sebastian @kuaashish @tyrmullen @kalyan2789g

### DIFF
--- a/mediapipe/tasks/web/audio/audio_classifier/BUILD
+++ b/mediapipe/tasks/web/audio/audio_classifier/BUILD
@@ -16,6 +16,7 @@ ts_library(
     visibility = ["//visibility:public"],
     deps = [
         ":audio_classifier_types",
+        "//mediapipe/calculators/tensor:tensors_to_classification_calculator_jspb_proto",
         "//mediapipe/framework:calculator_jspb_proto",
         "//mediapipe/framework:calculator_options_jspb_proto",
         "//mediapipe/tasks/cc/audio/audio_classifier/proto:audio_classifier_graph_options_jspb_proto",
@@ -26,6 +27,7 @@ ts_library(
         "//mediapipe/tasks/web/components/containers:classification_result",
         "//mediapipe/tasks/web/components/processors:classifier_options",
         "//mediapipe/tasks/web/components/processors:classifier_result",
+        "//mediapipe/tasks/web/components/processors:label_map",
         "//mediapipe/tasks/web/core",
         "//mediapipe/tasks/web/core:classifier_options",
         "//mediapipe/tasks/web/core:task_runner",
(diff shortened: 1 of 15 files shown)


marcusx2:
@a2105z

> classifier.getDisplayNames(); // ["manzana", "plátano"] or ["", ""]

Does this make sense? My understanding is that you either have display names for all categories or you don't have them at all.

Are you saying it's possible to get something like this?

> ["manzana", "", "", "plátano"]

Is that why the API returns an array of empty strings when display names aren't available, instead of returning something more appropriate like undefined, null, or an empty array?

a2105z:
@marcusx2 You're right — that example was misleading.

TFLite metadata is all-or-nothing. `BuildLabelMapFromFiles` either loads a locale associated file with **the same number of lines as the label file**, or it loads no display names at all. A mixed list like `["manzana", "", "", "plátano"]` is not something the metadata extractor produces.

`["", ""]` was the wrong signal. `getDisplayNames()` now matches `getLabels()`:

- model has a display-name file → full list, same length as `getLabels()`, index-aligned (`["manzana", "plátano"]`)
- model has none → `[]`

Not `undefined` / `null`: the rest of the Web Tasks API (including `getLabels()`) uses empty arrays for “absent”, so callers can write `if (classifier.getDisplayNames().length)`.

Pushed that change just now.

